PR #22, opened 2020-07-23
Not important, small simplifications and information

NikV2:
Simplification, small usage of lambdas, and information that is required for new users when going through the settings such as the doAutoResolveClientVersion


retrooper:
I intentionally didn’t like using lambda to support legacy java versions, and I am just used to having the whole thing visible.
I see you out in effort, I’ll accept it and try to use lambda more aswell!

NikV2:
I never liked the usage of lambdas either, Thing is, the java version we're aiming for in our case. It doesnt really matter if we do or do not use lambdas. I'm not saying you *Should* use lambdas, it's just little things like the one i changed, where in my opinion it makes more sense to use a lambda. Thanks for accepting!

PR #28, opened 2020-08-27
Small changes

NikV2:
Added an option to choose whether or not Uninjection should be handled Async or not (false by default), Removed the option to set the identifier since now it generates a random string using the alphabet (Not sure if this is how it should be handled, Please let me know!), The listeners now have the Lowest priority, Giving the option for other plugins to customize the outcome(Not important)


retrooper:
Small???? How is this small? This is perfect, thanks a lot for the contribution!

NikV2:
Thank you! Seems like i forgot to remove the method that i deleted in the example class, Hence why compiling fails

PR #29, opened 2020-08-27
Remove committer from javadoc workflow

SamJakob:


SamJakob:
(The check failed because PRs don't have authorization to deploy.)

PR #37, opened 2020-09-13
Optimized reflection, fixed WrappedPacketInFlying error

funkemunky:
- Got rid of the MethodHandle usage. It isn't the LambdaMetaFactory, and it's constantly creating invoke dynamics which are fairly heavy. Just put standard reflection there instead (12000ns vs now 800ns average). 
- Removed all the individual WrappedInFlyingPacket wrappers. Not necessary, and they don't make code much cleaner.
- Fixed the packetClass code execution ordering. I'm willing you bet you were scratching your head on why it wasn't grabbing parent classes properly, but here you are. Also, I don't recommend you just constantly grab fields. You should cache the classes and fields statically in a Map or something. That'll add an even larger performance boost RAM and CPU wise since the garbage collector will do less work and you're not running CPU instructions for the same task with the same result repeatedly. .


retrooper:
really appreciate the work you have done, join my discord server for some role!

PR #45, opened 2020-09-20
You now have to provide a Settings class when starting PE

GoDead:


retrooper:
Pull requests should be sent to the dev branch.

PR #48, opened 2020-09-26
Small change - Re-did this to possibly fix the conflicts the previous PR had

NikV2:
Fixes invocationtargetexception caused rarely in unknown situations or when the server's lagging like hell and the player leaves suddenly and/or reloading (strange, also not really a fix but more like a bandaid)

And it also looks a bit cleaner? (Yes i love lambdas)

### DIFF
--- a/src/main/java/io/github/retrooper/packetevents/PEVersion.java
+++ b/src/main/java/io/github/retrooper/packetevents/PEVersion.java
@@ -44,9 +44,6 @@ public class PEVersion {
             this.version[1] = version[1];
             if (len == 3) {
                 this.version[2] = version[2];
-            } else {
-                this.version[2] = 0;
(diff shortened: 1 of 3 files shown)


retrooper:
The PEVersion redundant initiation was already fixed in the dev branch before you made this pull request.
The EventSynchronization system only called the Bukkit Scheduler 'runTaskAsynchronously' method to call your method asynchronously.
To call your method synchronously, PacketEvents used the 'runTask' synchronously method in the Bukkit Scheduler.

I think it is much better you just call it yourself, inside your method whenever you need it. So that change is also denied.

That local invokeMethod runnable variable was redundant, never used as the eventsynchronization system has been removed, so your lambda usage change has also been denied.

PR #50, opened 2020-09-27
packetevents 1.7 pre release 1(1.6.9.1)

retrooper:
#### The first pre release for 1.7

* All clientside wrappers have been optimized to only read the fields you access through the getters(performance improvement)
* A few wrapper bug fixes
... A LOT MORE, BE READY FOR 1.7 RELEASE

Remember, this is a pre release, it **MIGHT** not be stable.
It is recommended to test pre releases, and not trust them to always work.
I do accept bug reports on pre release versions, just not on dev builds.


retrooper:
1.7's first pre release! 😄

PR #52, opened 2020-10-07
added isBungeecord because i want to use it ty

LIWKK:


retrooper (changes requested):
Rather have this in ServerUtils class if it is publicly used, good idea.
Anyway I need to make a whole new Bungee cord version of PE to support BUNGEE.
Thats why I don’t do anything if it is a bungee version.

retrooper:
Your request done in this *dev* commit: https://github.com/retrooper/packetevents/commit/ac585f7c81b34caee2f87961dfd84babd32a8d19

PR #73, opened 2020-12-02
Minor Improvements & Code Cleanup

NikV2:
Cleaned up imports
Minor code changes
Added a small check to make sure the users do not go above the thread amount of the server


NikV2:
Conflicts, god damn.

retrooper (changes requested):
Everything seems alright, except for one thing. 
Runtime.getRuntime().availableProcessors() returns the amount of cores available to the JVM.
There are two issues with that part of code you pull requested.
Let us say I have 2 cores for example.
That does not mean I am limited to 2 threads otherwise the system will break.
It is recommended to use two threads per core, resulting in 2 * 2 = 4.
Minecraft(and most other games) would allocate 4 threads for their networking for example.(Packet processing, connection handling, etc.)
What the core will do is it will switch threads.
If I would allocate 12 threads, it could still run alright, just might end up performing worse than 4.
Switching threads does have a performance cost, so I really am careful when doing allocated thread count checks.
I assume whoever is setting the thread count knows what they are doing. The default value is 1, which most people using PacketEvents have tested to run fine on 100-200 players.
The main reason it is set to 1 is to avoid as many concurrency issues with multiple threads accessing and modifying data.

NikV2:
Thank you!
I actually didn't know that, I'll definitely use that *Thread Formula* for future reference.

NikV2:
As far as i know when multithreading, It's usually much safer to stay within the Runtime.getRuntime.availableProcessors value. So i figured i'd make a check for it on startup.

retrooper:
> As far as i know when multithreading, It's usually much safer to stay within the Runtime.getRuntime.availableProcessors value or lower. So i figured i'd make a check for it on startup.

Try doing your own research and depend less on me.
I could be lying @NikV2

NikV2:
Would you like me to remove my check on startup and re-open?

retrooper:
> As far as i know when multithreading, It's usually much safer to stay within the Runtime.getRuntime.availableProcessors value or lower since that returns the optimal value. So i figured i'd make a check for it on startup.

afaik Runtime.getRuntime.availableProcessors * 2 is the optimal value.

NikV2:
Should we really allow users to go that high though?:|
I don't think it'd turn out being efficient if a user uses that many threads for that.

Feel free to let me know if u'd like me to modify it or completely remove it.

NikV2:
According to a random wiki i just found it returns the number of processors available to the JVM
https://www.tutorialspoint.com/java/lang/runtime_availableprocessors.htm

retrooper:
> According to a random wiki i just found it returns the number of processors available to the JVM which should basically mean that you usually want to avoid going higher than that.
> https://www.tutorialspoint.com/java/lang/runtime_availableprocessors.htm

The number of processes is the number of cores. 
Here is what a core is: https://www.tomshardware.com/news/cpu-core-definition,37658.html

retrooper:
>. .. the number of processors available to the JVM which should basically mean that you usually want to avoid going higher than that.

Not necessarily...
You can even use multiple threads on one of those old SINGLE CORE CPUs.
https://stackoverflow.com/questions/16116952/can-multithreading-be-implemented-on-a-single-processor-system#:~:text=Yes%20you%20can%20do%20multithreading,execute%20%2C%20simultaneously%20on%20different%20cores.&text=But%20the%20running%20threads%20%2C%20gives,of%20the%20User%20space%20application.

PR #100, opened 2021-02-20
Fix 1.7.10 loadup error

LIWKK:


retrooper:
Thanks a ton, some contributor accidentally messed up when tryna clean that up.

PR #105, opened 2021-02-23
Method to check if player is inject in handlers.

LIWKK:


retrooper (approved):
Great feature. Thanks for the PR!

PR #121, opened 2021-05-24
Added a suggested wrapper

Tecnio:
Added WrappedPacketOutBlockAction as a readable and sendable wrapper.


retrooper (changes requested):
I think WrappedPacketOutBlockAction#getBlockType would be a better name rather than WrappedPacketOutBlockAction#getBlockMaterial

PR #122, opened 2021-05-25
Added 2 wrappers.

Tecnio:
Added WrappedPacketInBEdit and WrappedPacketInJigsawGenerate.


retrooper (changes requested):
If possible setters for the wrappers. Can you explain why WrappedPacketInBEdit is deprecated? I think the getHand method in the BEdit wrapper can be cleaned up using my enum utilities. Look at how I implemented a getHand on my WrappedPacketInArmAnimation and can just use java's Optional class. Seems like the hand is only accessible on 1.13 -> 1.16.5, so on those versions return Optional.of(value), otherwise Optional.empty(). This is a bit more complicated, so you can just comment out your hand method and I can finish it :)

retrooper (approved):
Thank you for the wrappers and for the optimization/cleanup in arm animation.

PR #123, opened 2021-05-25
Made 2 login packets sendable

Tecnio:
Made WrappedPacketLoginOutDisconnect and WrappedPacketLoginOutSuccess sendable.


retrooper (changes requested):
@Override    public Object asNMSPacket() throws Exception {        return packetConstructor.newInstance(                NMSUtils.generateIChatBaseComponent(this.reason)        );    }

should be

@Override    public Object asNMSPacket() throws Exception {        return packetConstructor.newInstance(                NMSUtils.generateIChatBaseComponent(getReason())        );    }

Imagine the case someone listens to a packet and they want to send the same packet. The reason variable would be null. Can you fix this in the other wrapper too? And if I missed, in the past wrappers you made too?

PR #124, opened 2021-06-03
Remove assumption of SocketChannel

Camotoy:
Hello, Geyser person here. We're working on some changes to inject Geyser players in the server so they don't create a TCP connection. However, the channels they use aren't SocketChannels, meaning this library would break with the changes. This PR rectifies that by setting all mentions of SocketChannel to Channel. Tested with Paper 1.16.5 and CraftBukkit 1.7.10 using the Attack Detector example plugin.

Without these changes, https://github.com/PaperMC/Paper/pull/5611 will also break. 

Thanks!

### DIFF
--- a/src/main/java/io/github/retrooper/packetevents/injector/legacy/early/EarlyChannelInjectorLegacy.java
+++ b/src/main/java/io/github/retrooper/packetevents/injector/legacy/early/EarlyChannelInjectorLegacy.java
@@ -29,7 +29,6 @@ import net.minecraft.util.io.netty.channel.Channel;
 import net.minecraft.util.io.netty.channel.ChannelFuture;
 import net.minecraft.util.io.netty.channel.ChannelHandler;
 import net.minecraft.util.io.netty.channel.ChannelInitializer;
-import net.minecraft.util.io.netty.channel.socket.SocketChannel;
 import org.bukkit.entity.Player;
 import org.bukkit.plugin.PluginDescriptionFile;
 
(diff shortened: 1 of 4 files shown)


Camotoy:
I would recommend that you remove assumptions that the channel remote address is an INetSocketAddress, such as here: https://github.com/retrooper/packetevents/blob/dev/src/main/java/io/github/retrooper/packetevents/utils/netty/channel/ChannelUtils8.java which will break if anyone uses the setting present in that PR.

retrooper:
I am honored by your contribution Cametoy. I have been considering working on Geyser support, I appreciate your help. Why can’t one assume the channel is an INetSocketAddress?

Camotoy:
https://netty.io/4.1/api/io/netty/channel/epoll/EpollDomainSocketChannel.html#remoteAddress-- which is what that PR will return, which is incompatible with INetSocketAddress.

retrooper:
How can I identify users with such a channel? Can I just use the channel object, does it contain some unique data. When I am processing early packets like the SET PROTOCOL one, I do not have access to the user’s UUID yet. So I will store the protocol versions in a map with the inetaddress as key and clientversion as value. Are these channels unique in anyway?

Camotoy:
I'm afraid that is beyond my knowledge - both in terms of a better implementation and how domain socket addresses work. All I can say is that code could break in the future.

retrooper:
So when accessing the remote address(getting inetsocketaddress) break? Im a bit confused about the heads up. You said in future code it will work?

Camotoy:
Don't assume that `channel.remoteAddress()` will return an INetSocketAddress, is my recommendation. But this is beyond the scope of this PR, and worst case you can address the problem when someone actually tries to make use of that PR.

Camotoy:
Thank you! I really appreciate it.

retrooper:
I'll look into not using INetSocketAddress, I appreciate you going out of your way to contribute. I really support Geyser.

PR #138, opened 2021-07-09
Added WrappedPacketOutCollect

Tecnio:
As the title suggests added a wrapper for PacketPlayOutCollect aka Collect Item


retrooper:
On 1.17, the correct constructor is at index 0, not index 1.

retrooper:
If a field is not supported on 1.7.10 -> 1.17, we use Optionals. We do that to make it clear that a field might not be supported on every version.

retrooper:
```java
 return packetConstructor.getParameterCount() == 3
                ? packetConstructor.newInstance(this.collectedEntityId, this.collectorEntityId, this.itemCount)
                : packetConstructor.newInstance(this.collectedEntityId, this.collectorEntityId);
```

When doing this, please use the getters for these variables, and not the variable itself. Not doing this can cause bugs. I've already told you this. Lets have a look at a getter in the wrapper.
```java
public int getCollectedEntityId() {
        if (packet != null) {
            return readInt(0);
        } else {
            return this.collectedEntityId;
        }
    }
```
You see, the packet object can also be not null, meaning this packet is being sent by the server. The user is trying to read the packet that is outgoing. If packet is null, the developer is trying to create a packet with this wrapper. Now when the packet is not null, the collectedEntityId variable is  unused. Which means it will hold the value 0. 

In simple words, you should do this:

 ```java
 return packetConstructor.getParameterCount() == 3
                ? packetConstructor.newInstance(getCollectedEntityId(), getCollectorEntityId(), getItemCount())
                : packetConstructor.newInstance(getCollectedEntityId(), getCollectorEntityId());
```

Issue #136, opened 2021-07-02
PacketType.Play.Client.FLYING ID broken on 1.17


MWHunter:
This packet is the position packet with onGround status only.  You can get this packet by towering upwards only pressing space and right click, not looking or moving in any other way.  The packet works on 1.16 but on 1.17, the ID is -128.

if (packetID == PacketType.Play.Client.FLYING) {

MWHunter:
Broken on both spigot and paper 1.17

retrooper:
?

retrooper:
PacketType.Play.Client.FLYING isn't broken.

retrooper:
-128 Packet ID means its an undocumented/unknown Packet ID. I have double-checked, I see no issue with FLYING.

MWHunter:
The problem is that this is a packet ID that exists on 1.17.  For some reason, I can check when the packet arrives with this code:

if (event.getPacketName().equalsIgnoreCase("d")) {
}

In here, when you get the packet ID, it will be -128.  This is wrong because the packet exists but the packet ID believes it does not exist.

PR #142, opened 2021-07-10
Fix FLYING packet on 1.17

MWHunter:
It's a subclass of the FLYING packet and isn't a packet itself.  I'm calling it GROUND because it only has the ground status without anything else updated.  I've left the packet name as FLYING to maintain compatibility with previous versions, although it would be technically more correct to rename it all to GROUND.


MWHunter:
Fixes https://github.com/retrooper/packetevents/issues/136

Issue #139, opened 2021-07-10
1.17.1 issue with PacketPlayOutEntityDestroy


MWHunter:
> io.github.retrooper.packetevents.exceptions.WrapperFieldNotFoundException: PacketEvents failed to find a int indexed 0 by its type in the net.minecraft.network.protocol.game.PacketPlayOutEntityDestroy class!
        at io.github.retrooper.packetevents.packetwrappers.WrappedPacket.getField(WrappedPacket.java:508) ~[?:?]
        at io.github.retrooper.packetevents.packetwrappers.WrappedPacket.read(WrappedPacket.java:288) ~[?:?]
        at io.github.retrooper.packetevents.packetwrappers.WrappedPacket.readInt(WrappedPacket.java:194) ~[?:?]
        at io.github.retrooper.packetevents.packetwrappers.play.out.entitydestroy.WrappedPacketOutEntityDestroy.getEntityId(WrappedPacketOutEntityDestroy.java:89) ~[?:?]
        at io.github.retrooper.packetevents.packetwrappers.play.out.entitydestroy.WrappedPacketOutEntityDestroy.getEntityIds(WrappedPacketOutEntityDestroy.java:121) ~[?:?]
        at ac.grim.grimac.events.packets.PacketFireworkListener.onPacketPlaySend(PacketFireworkListener.java:61) ~[?:?]
        at io.github.retrooper.packetevents.event.impl.PacketPlaySendEvent.call(PacketPlaySendEvent.java:78) ~[?:?]
        at io.github.retrooper.packetevents.event.manager.EventManagerModern.callEvent(EventManagerModern.java:58) ~[?:?]
        at io.github.retrooper.packetevents.event.manager.EventManager.callEvent(EventManager.java:52) ~[?:?]
        at io.github.retrooper.packetevents.processor.PacketProcessorInternal.write(PacketProcessorInternal.java:163) ~[?:?]
        at io.github.retrooper.packetevents.injector.modern.PlayerChannelHandlerModern.write(PlayerChannelHandlerModern.java:46) ~[?:?]
        at io.netty.channel.AbstractChannelHandlerContext.invokeWrite0(AbstractChannelHandlerContext.java:717) ~[patched_1.17.1.jar:git-Paper-97]
        at io.netty.channel.AbstractChannelHandlerContext.invokeWriteAndFlush(AbstractChannelHandlerContext.java:764) ~[patched_1.17.1.jar:git-Paper-97]
[...]

PR #143, opened 2021-07-10
Fix reading destroy packet on 1.17

MWHunter:
Setting variables is still probably broken, but this fixes reading the destroy packet on 1.17.1.

Tested and works on 1.7, 1.17, and 1.17.1

I also removed some unused imports from the class

Setting variables is still probably broken but at least reading works with this commit.


MWHunter:
Fixes https://github.com/retrooper/packetevents/issues/139

retrooper:
I'll be committing some additions to fix some things you forgot to account for.

retrooper:
I'll be committing some things you forgot to account for.

PR #153, opened 2021-08-02
Adding PacketDecodeEvent

abhiram555:
So basically this event is added for my plugin to work. but I thought why not share it with community. you can use this event to check for BadPackets in anti-cheats or other crasher checks


retrooper:
Have you tested this to work on 1.7.10 and 1.8+?

abhiram555:
> Have you tested this to work on 1.7.10 and 1.8+?

Tested on 1.8 and 1.12 works fine

retrooper:
Since the PacketDecodeEvent you contributed is called after the bytebuf is decompressed, We might need to call it before the decompression. We can make a separate event for when its decompressed, for example PacketDecodeEventAfterDecompress

abhiram555:
> Since the PacketDecodeEvent you contributed is called after the bytebuf is decompressed, We might need to call it before the decompression. We can make a separate event for when its decompressed, for example PacketDecodeEventAfterDecompress

will add that.

abhiram555:
> Since the PacketDecodeEvent you contributed is called after the bytebuf is decompressed, We might need to call it before the decompression. We can make a separate event for when its decompressed, for example PacketDecodeEventAfterDecompress

added the packet decoder before minecraft's packet decoder so yeah PacketDecodeEventAfterDecompress is not needed since you could handle compressed packet your own (by listening for SetCompressionPacket)

abhiram555:
Added Decoder Before Minecraft's decoder in pipeline also added a getter for getting Refcnt in ByteBuffer

PR #156, opened 2021-08-09
Update ViaVersionAccessorImplLegacy.java

LIWKK:


retrooper:
Has already been fixed. We use reflection on older ViaVersion API to avoid some incompatibilities and exceptions many users have reported me before when I tried your approach. It now works, but I still recommend users to update their ViaVersion. I wish I could just depend on the new Via API, but its sad some people just don't update or some developers shade ViaVersion into their Spigot, then never update it...

PR #168, opened 2021-10-07
Fix packetevents multi-module issues :D

conclube:


retrooper:
Thanks for helping me with making packetevents a multi-module project.

PR #172, opened 2021-12-02
Update gradle shadow & add github actions to 2.0

AlessioGr:


retrooper (changes requested):
Build with Java 8

MWHunter:
It shouldn't matter if you build it with Java 17.  It's going to work fine on java 8 through 17, following what's specified in gradle.  Actually, it's best to compile with latest as new java versions are more strict about hacky reflection than old versions.

PR #173, opened 2021-12-11
Add LogManager

AlessioGr:
Signed-off-by: AlessioGr <alessio@gravili.de>


retrooper:
What is the plugin.yml.dis?

AlessioGr:
@retrooper 

Well in case you wanna compile a plugin jar and test it you can just rename the .jar.dis to .jar (.dis stands for disabled). Better than making one from scratch or constantly deleting it.

As you said, it causes issues keeping the plugin.yml there the entire time - that's what the .dis is for.

retrooper:
Great idea.

retrooper (approved):

PR #174, opened 2021-12-12
Add vanilla MinecraftEncryptionUtil.

TRGReal:
Allows you to easily encrypt/decrypt using Minecraft-like RSA/AES in PE util.


retrooper:
Thanks, and nice documentation.

PR #197, opened 2022-01-19
Adventure implementation

AbhigyaKrishna:
Adventure implementation for packet wrappers. The serialization of adventure component is platform independent.

### DIFF
--- a/api/build.gradle
+++ b/api/build.gradle
@@ -13,6 +13,8 @@ tasks {
 dependencies {
     compileOnlyApi 'com.google.code.gson:gson:2.8.9'
     compileOnly 'org.jetbrains:annotations:22.0.0'
+    implementation 'net.kyori:adventure-api:4.9.3'
+    implementation 'net.kyori:adventure-text-serializer-gson:4.9.3'
 }
 
 publishing {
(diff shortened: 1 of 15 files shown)


MWHunter:
fine by me if it works

retrooper:
Noticed you try to keep our current chat abstraction, you don't need to. Adventure chat api can just replace our component system honestly. Thanks a lot @AbhigyaKrishna and I will do a few minor changes and some tests.

AbhigyaKrishna:
I wasn't sure if i should remove the current chat serialization so i just implemented adventure on top of it

MWHunter:
If we use implementation we need the shading plugin otherwise adventure won't actually be in the jar

retrooper:
That is handled.

MWHunter:
I did ./gradlew build and it didn't shade adventure into the jar

retrooper:
Its not shaded into the API, its into Spigot jar. You are checking the api jar, aren't you?

retrooper:
I'll see what I can do to get the libraries in the API module, cause I remember running into issues when trying to shade it into the API module, then shade the API module into spigot.

MWHunter:
Spigot doesn't shade adventure, only paper does on 1.16+ or a version around there

retrooper:
I know.

PR #198, opened 2022-01-20
Added WrapperPlayServerScoreboardObjective and WrapperPlayServerTeams

RohanGoyalDev:


MWHunter:
Shouldn't LegacyComponentSerializer.parseChar() and other Adventure methods be used over adding things to ColorUtil?  If anything, ColorUtil should be entirely removed as Adventure should have all this functionality already

retrooper:
> Shouldn't LegacyComponentSerializer.parseChar() and other Adventure methods be used over adding things to ColorUtil? If anything, ColorUtil should be entirely removed as Adventure should have all this functionality already

Hoped they had such functionality.

retrooper:
Not sure parseChar is what we need anywhere?

PR #199, opened 2022-01-21
Added unregisterListeners methods

RohanGoyalDev:


retrooper:
I don't think I can accept this PR, unregistering a specific listener instance in 2.0 would be tough since we have various packet listener classes. PacketListener, PacketListenerReflect, PacketListenerAbstract. They are all at the end of the day converted to the abstract one, so you couldn't unregister those ones.

RohanGoyalDev:
What if we make the register methods return PacketListenerAbstract instead of void

retrooper:
Thats a good idea, I'll look into it tomorrow.

PR #218, opened 2022-02-02
Fixes for legacy mappings

AbhigyaKrishna:
Some mapping changes for legacy PacketPlayOutSpawnEntity


retrooper:
Whats point of the getLegacyId, can't it just be getId?

AbhigyaKrishna:
For some reason they use different ids in PacketPlayOutSpawnEntity packet in versions 1.13 and below
image

and above that they are using the entity ids as defined.

retrooper:
I'm not 100% following, but can't you then just override getId for legacy versions too? So its easier and we only use one method for ids? Or am I missing something?

MWHunter:
I think he is talking about how living entities and non-living entities use the same set of overlapping ID's.  If this is the case, I'd suggest just combining both living and non-living into the same map.  I've included a helper method in EntityType to tell you if some entity is considered alive (armor stands are alive! so it isn't intuitive)

AbhigyaKrishna:
According to the previous code which packetevent had it was writing the id of the entity as in the mappings. Yes this will work for SpawnLivingEntity packet coz the SpawnEntityLiving packet uses the same ids as of the entity id which are mapped already. But for some reason in versions below 1.14 the ids in SpawnEntity packet (not SpawnEntityLiving) it uses different byte mappings than the entity ids corresponding to the entity.

image

For, an example,
I was trying to spawn a dropped item which has a entity id of 1 but it didnt work because the SpawnEntity packet used byte 2 to spawn a dropped item so i had to use EntityType.EXPERIENCE_ORB as it has the id 2.

And for the items mapping part i clearly sent the metadata for a stone item but it spawned a coal item as the entity id of coal in 1.8 is 16 which is similar to that of stone for the mapping which was provided.

retrooper:
@AbhigyaKrishna What do you think about naming the method getNonLivingId() and for modern versions, it just also returns getId?

AbhigyaKrishna:
Not sure if we should do that as not all entities have their non-living id iirc. And it's up to you what you would be naming that as it is only used internally.

retrooper:
Lets keep it as getLegacyId.

retrooper:
Is 1.7 supported?

AbhigyaKrishna:
Yes, 1.7.10 is same as 1.8

PR #231, opened 2022-02-11
Use Mojang Mappings with ParchmentMC

Olivoz:
Use Mojang Mappings with ParchmentMC so method names are the same across different platforms.


retrooper:
Thanks.

PR #260, opened 2022-02-23
Version bump for adventure to use snapshot

AbhigyaKrishna:


retrooper:
Do you think this will fix the adventure issues for legacy versions?

AbhigyaKrishna:
Yeah this might fix it because they have many internal changes for parsers too

retrooper:
Why does GitHub actions fail to build?

derklaro:
Because the spigot sub project is unable to resolve the dependencies of the api project... The snapshot repo should be declared in the parent build file I gues...

derklaro:
FYI they just release 4.10

retrooper (changes requested):
Update to 4.10

retrooper:
https://github.com/retrooper/packetevents/commit/65bc4e18b032150a139a66710696baaa28172995

PR #284, opened 2022-03-12
WrapperPlayServerResourcePackSend

amnoah:
I think I did it correctly, please let me know if anything is wrong.

Basically copied from MCProtocolLib

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerResourcePackSend.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerResourcePackSend.java
@@ -19,6 +19,7 @@
 package com.github.retrooper.packetevents.wrapper.play.server;
 
 import com.github.retrooper.packetevents.event.PacketSendEvent;
+import com.github.retrooper.packetevents.manager.server.ServerVersion;
 import com.github.retrooper.packetevents.protocol.packettype.PacketType;
 import com.github.retrooper.packetevents.wrapper.PacketWrapper;
 import net.kyori.adventure.text.Component;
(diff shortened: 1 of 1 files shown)


api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerResourcePackSend.java
```diff
@@ -0,0 +1,88 @@
...
+        prompt = wrapper.prompt;
+    }
+
+    @Override
+    public void write() {
+        writeString(url);
+        writeString(hash);
+        writeBoolean(required);
+        writeBoolean(prompt != null);
+        writeComponent(prompt);
```
MWHunter, line 70: prompt isn't written if it is null

api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerResourcePackSend.java
```diff
@@ -0,0 +1,88 @@
...
+        this.hash = hash;
+        this.required = required;
+        this.prompt = prompt;
+    }
+
+    @Override
+    public void read() {
+        url = readString();
+        hash = readString();
+        required = readBoolean();
```
MWHunter, line 48: https://wiki.vg/index.php?title=Protocol&oldid=16681#Resource_Pack_Send

1.16 and below doesn't have the required or prompt

MWHunter (changes requested):

api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerResourcePackSend.java
amnoah: Fixed in latest commit, Required/hasPrompt/Prompt will no longer be accessed unless the server is above or equal to version 1.16.

api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerResourcePackSend.java
amnoah: Fixed this by adding "hasPrompt" boolean (following Wiki.vg names)

MWHunter:
Could be better with another constructor for pre-1.17 servers but I really don't care and that can be added later when we figure out how to make better constructors that don't require every argument

PR #310, opened 2022-04-07
Fixed incorrect HealthDisplay data in ScoreboardObjective wrapper

LeeGodSRC:
ScoreboardObjective wrapper wasn't using the correct health display read/write for 1.13+

This PR fixes it


retrooper:
Can we also merge this into version-dynamic branch?

PR #319, opened 2022-04-11
Don't clone on vector add/subtract

Jaren8r:
Right now vectors are this super weird mix of mutable/immutable, with the add/subtract methods cloning the vector before doing their operation. I think it would be better if these methods operated on the object itself instead of cloning and returning, as vectors aren't immutable and current behavior is weird.


MWHunter:
This is what bukkit does with vectors.  This is the opposite of what vanilla does with vectors.

Jaren8r:
Still would be best to pick a side on whether to be mutable or immutable, not have the worst of both worlds. I think mutable is the better choice here as you can always use .clone to get a vector you're sure nobody else will change.

retrooper:
Yes

PR #329, opened 2022-04-15
Make vectors immutable

Jaren8r:
Since making vectors fully mutable was a pretty unpopular change, this PR makes them fully immutable.


retrooper (approved):
Thanks for the changes.

PR #332, opened 2022-04-16
Add .with[XYZ] to vectors

Jaren8r:
This makes it easy to clone a vector changing only one of the values in one line.

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/util/Vector3d.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/util/Vector3d.java
@@ -168,6 +168,22 @@ public class Vector3d {
         return this.x * other.x + this.y * other.y + this.z * other.z;
     }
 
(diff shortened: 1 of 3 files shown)


Izibane:
wouldn't it better calling it **clone** then rather than **with**?

MWHunter:
Vectors are immutable, you can't change X, Y, and Z once they are set.  They are final fields.

Jaren8r:
> wouldn't it be better calling it **clone** then rather than **with**? you could also override the existing clone method

While I agree that **with** may not be the most clear name, I don't really agree with **clone** as this is doing more than just cloning it. it's making a new one with a different X/Y/Z value, which becomes useful when you're chaining methods on vectors and you want to make a copy of a vector **with** a different x/y/z (since you can't do `.clone().setY(10)` anymore, the idea is that you can do `.withY(10)` and get that behavior back without ruining the chain)

MWHunter:
lgtm

PR #349, opened 2022-05-08
Use compression threshold set in server.properties

booky10:


MWHunter:
what versions did you test this on?

booky10:
Tested in 1.18.2 and fixed it for 1.8, which uses java properties.

retrooper:
We already discussed this in the discord. Sorry to deny this, but I just came up with a better solution. Our packet compression utils just recreate compression when we can just use vanilla's compression handlers with reflection. These compression operations aren't called often, just like once or twice per user once the server adds the compression handlers.

PR #360, opened 2022-05-09
Added WrapperPlayClientSetDisplayedRecipe

NoJokeFNA:
Added `WrapperPlayClientSetDisplayedRecipe` wrapper class.
Tested Server-Versions: 1.18.2

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/wrapper/play/client/WrapperPlayClientSetDisplayedRecipe.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/wrapper/play/client/WrapperPlayClientSetDisplayedRecipe.java
@@ -1,0 +1,59 @@
+/*
+ * This file is part of packetevents - https://github.com/retrooper/packetevents
+ * Copyright (C) 2021 retrooper and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


api/src/main/java/com/github/retrooper/packetevents/wrapper/play/client/WrapperPlayClientSetDisplayedRecipe.java
```diff
@@ -0,0 +1,59 @@
...
+ */
+
+package com.github.retrooper.packetevents.wrapper.play.client;
+
+import com.github.retrooper.packetevents.event.PacketReceiveEvent;
+import com.github.retrooper.packetevents.protocol.packettype.PacketType;
+import com.github.retrooper.packetevents.wrapper.PacketWrapper;
+
+public class WrapperPlayClientSetDisplayedRecipe extends PacketWrapper<WrapperPlayClientSetDisplayedRecipe> {
+    private String recipe;
```
MWHunter, line 26: Recipe should be a ResourceLocation not a String

"minecraft:stick"
It makes getting the namespace, minecraft, separated from the value, stick, much easier

MWHunter (changes requested):

retrooper:
Seems like it was added in 1.16.2.

PR #380, opened 2022-05-11
Added WrapperPlayServerSculkVibrationSignal

NoJokeFNA:
https://wiki.vg/Protocol#Sculk_Vibration_Signal
Tested Server-Versions: 1.18.2


retrooper (changes requested):
Instead of a string identifier, maybe see if the ResourceLocation can be used. PacketWrapper#readResourceLocation or readIdentifier, is what it might be called in the wrappers.

NoJokeFNA:
I will test it today or tomorrow.

retrooper:
Does it work on 1.8?

NoJokeFNA:
No. Sculks were introduced on newer versions.

retrooper:
On wiki.vg it says the "destination" (field after the destination identifier) can either be an ENTITY ID or an encoded block position. So this code might not always work? https://wiki.vg/Protocol#Sculk_Vibration_Signal

retrooper (changes requested):
On wiki.vg it says the "destination" (field after the destination identifier) can either be a VarInt ENTITY ID or a Long encoded block position. So this code might not always work? https://wiki.vg/Protocol#Sculk_Vibration_Signal

NoJokeFNA:
Yea but I don't know under which circumstances. I tested it for a few time and haven't received any errors or such.

retrooper:
Have you checked minecraft's vanilla code?

retrooper:
I'll check.

retrooper:
Depends on the identifier. I think the identifier can either be "minecraft:entity" or "minecraft:block" according to vanilla code.

retrooper:
But I'd say do some debugging too.

NoJokeFNA:
Alright, will do it later

PR #382, opened 2022-05-12
Added WrapperPlayServerUnlockRecipes

NoJokeFNA:
https://wiki.vg/Protocol#Unlock_Recipes (1.14 -> 1.18.2)
https://wiki.vg/index.php?title=Protocol&oldid=14299 (1.13)
https://wiki.vg/index.php?title=Protocol&oldid=12930 (1.12)
Tested Server-Versions: 1.12, 1.12.1, 1.12.2, 1.13.2, 1.14, 1.14.4, 1.15.2, 1.16.4, 1.18.2
~~Current issue
I have some problems with server versions 1.14 to 1.15.2 always throwing an `IndexOutOfBoundsException` even though nothing has changed in those versions, but I think there is a general problem with the versions and it is not the code's fault.~~

~~The Client-Version doesn't matter in this case, I've tried it with multiple ones -> same issue.~~

~~Logs
https://sourceb.in/gNedCXvso3~~

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerUnlockRecipes.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerUnlockRecipes.java
@@ -253,10 +253,20 @@ public class WrapperPlayServerUnlockRecipes extends PacketWrapper<WrapperPlaySer
         this.elements = elements;
     }
 
+    /**
+     * On version 1.13 and above, the recipe ids are read as ResourceLocation, otherwise they are read as integers.
+     *
+     * @return The recipe ids
+     */
     public List<RecipeId<?>> getRecipeIds() {
         return recipeIds;
     }
 
+    /**
+     * On version 1.13 and above, the recipe ids are written as ResourceLocation, otherwise they are written as integers.
+     *
+     * @param recipeIds The recipe ids
+     */
     public void setRecipeIds(List<RecipeId<?>> recipeIds) {
         this.recipeIds = recipeIds;
     }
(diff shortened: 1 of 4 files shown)


api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerUnlockRecipes.java
```diff
@@ -0,0 +1,331 @@
...
+    } else {
+      this.action = Action.VALUES[readShort()];
+    }
+    this.craftingRecipeBookOpen = readBoolean();
+    this.craftingRecipeBookFilterActive = readBoolean();
+    if (serverVersion.isNewerThanOrEquals(ServerVersion.V_1_13)) {
+      this.smeltingRecipeBookOpen = readBoolean();
+      this.smeltingRecipeBookFilterActive = readBoolean();
+    }
+    if (serverVersion.isNewerThanOrEquals(ServerVersion.V_1_14)) {
```
Jaren8r, line 117:
```suggestion
    if (serverVersion.isNewerThanOrEquals(ServerVersion.V_1_16_2)) {
```
These properties were added in 1.16.2, not 1.14

Jaren8r (changes requested):

retrooper:
Interesting how you have modern and legacy recipe "inits"? I don't know if its possible to bring them into one form? If possible I don't want different fields available for different versions.

NoJokeFNA:
I know that, but I honestly don't know how we could make it different. We could use Objects, but you don't really wanna have Objects.

Axionize:
Use a Wrapper object inside of the WrapperPlayServerUnlockRecipes for the field instead of an `Object`. Make it accessible via getters/setters and abstract away the per-version logic

PR #389, opened 2022-05-15
Added WrapperPlayServerNBTQueryResponse

NoJokeFNA:
https://wiki.vg/Protocol#NBT_Query_Response
Tested Server-Versions: 1.16.5, 1.18.2


retrooper (approved):
Looking good. Maybe we can get someone else to approve it too actually checking if its valid.

PR #397, opened 2022-05-19
Fixed code-style

NoJokeFNA:


retrooper:
Make the getter of the angle a **float** also.

PR #398, opened 2022-05-19
Fixed code-style

NoJokeFNA:


retrooper:
Please look at the changes you made, it's changing an already merged & modified wrapper. You're reversing my changes.

NoJokeFNA:
Will just do another with your current changes as it might probably be the best option anyway, same for #397 (so you will only have one PR for the main-branch stuff)

PR #401, opened 2022-05-20
Added WrapperPlayServerSelectAdvancementTab

NoJokeFNA:
https://wiki.vg/Protocol#Select_Advancement_Tab
Tested Server-Versions: 1.12, 1.12.1, 1.18.2


retrooper:
In my opinion, the hasId boolean is redundant. You can just have the field be remain nullable, in the read method read the boolean and if true, set the identifier. while encoding you can set the boolean to true if the identifier is not null. and make a getter for the identifier remain an Optional.

retrooper (approved):
Looking good.

PR #428, opened 2022-06-11
[1.19] Fix sound wrappers

NoJokeFNA:
WrapperPlayServerNamedSoundEffect doesn't exist yet, so nothing can be fixed there


api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerEntitySoundEffect.java
```diff
@@ -50,6 +59,9 @@ public void read() {
         this.entityId = readVarInt();
         this.volume = readFloat();
         this.pitch = readFloat();
+        if (serverVersion.isNewerThanOrEquals(ServerVersion.V_1_19)) {
+            this.seed = readLong();
```
MWHunter, line 63: https://wiki.vg/Protocol#Entity_Sound_Effect

is protocol wiki wrong?  I don't see this

MWHunter (changes requested):

api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerEntitySoundEffect.java
NoJokeFNA: No you are right, sleepy eyes

retrooper:
For us it's been a "convention" to keep method names for IDs ending with getId() instead of getID()

retrooper:
So your change would make it inconsistent.

retrooper:
Seems like the OptionalLong is useless as you never make it empty, its either null or has a value?

NoJokeFNA:
But why would you name your field *ID* and your Getter/Setter *Id*? This is very illogical and contradictory for me

retrooper:
True, variable naming is contradictory, maybe we'll change that but lets keep the methods the same.

retrooper:
I don't see the point of the OptionalLong tho. You're right its never null.

NoJokeFNA:
Well, Optionals exist for one reason, to check if the value does exist.

retrooper:
In your case it always exists and is never empty is my point.

NoJokeFNA:
Changed it, you are right

PR #432, opened 2022-06-11
Fxed memory leak with UUID objects

NoJokeFNA:
Creating new UUID objects will lead to a memory leak after some time, so we just cache one value and use that.


MWHunter:
memory allocations != memory leak

api/src/main/java/com/github/retrooper/packetevents/util/UUIDUtil.java
```diff
@@ -22,6 +22,7 @@
 import java.util.regex.Pattern;
 
 public class UUIDUtil {
+    public static final UUID DUMMY = UUIDUtil.DUMMY;
```
MWHunter, line 25: isn't this referring to itself?

MWHunter (changes requested):

Issue #440, opened 2022-06-23
Unable to join server


scarfacered:
**Describe the bug**
https://pastebin.com/raw/TwXTj5qX

**Software brand**
Paper 1.19

**Plugins**
ViaVersion, ViaBackwards

**How To Reproduce**
Connect to 1.19 server

scarfacered:
I believe this caused it
https://github.com/retrooper/packetevents/commit/5f49f63e8fad548ebeb5ef7a58a2bacb1d7644c5#diff-6e61ed83ab8d18aa7c13046c2cb4db8354217fb1ef7b1745b5f4f8227cbef7e9R55

scarfacered:
> The size is complete different so it wouldn't make that much sense

https://github.com/retrooper/packetevents/blob/5f49f63e8fad548ebeb5ef7a58a2bacb1d7644c5/api/src/main/java/com/github/retrooper/packetevents/wrapper/PacketWrapper.java#L367

It's multiplied by 4 here? 255 * 4 = 1020, which is what you're seeing in that error log. I'm not sure how my serverAddress is longer than 255 letters, though. It's behind Velocity? I'd be curious to see the string it's trying to parse.

retrooper:
So is this issue resolved?

retrooper:
I'm gonna mark it as solved then...

PR #441, opened 2022-06-24
Bump max length for serverAddress

scarfacered:
After this commit I can no longer connect to my server:
https://github.com/retrooper/packetevents/commit/5f49f63e8fad548ebeb5ef7a58a2bacb1d7644c5#diff-6e61ed83ab8d18aa7c13046c2cb4db8354217fb1ef7b1745b5f4f8227cbef7e9R55

I'm behind TCPShield and Velocity so one of those might be the cause of big server addresses... but I'm not entirely sure. 

Fixes https://github.com/retrooper/packetevents/issues/440

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/wrapper/PacketWrapper.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/wrapper/PacketWrapper.java
@@ -372,8 +372,8 @@ public class PacketWrapper<T extends PacketWrapper> {
         } else {
             String s = ByteBufHelper.toString(buffer, ByteBufHelper.readerIndex(buffer), j, StandardCharsets.UTF_8);
             ByteBufHelper.readerIndex(buffer, ByteBufHelper.readerIndex(buffer) + j);
-            if (s.length() > maxLen) {
-                throw new RuntimeException("The received string length is longer than maximum allowed (" + j + " > " + maxLen + ")");
+            if (s.length() > maxLen * 4) {
+                throw new RuntimeException("The received string length is longer than maximum allowed (" + s.length() + " > " + maxLen + ")");
             } else {
                 return s;
             }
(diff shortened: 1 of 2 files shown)


MWHunter:
Where did you get 350 from?

NoJokeFNA:
Yea it's trug that tcpshield increases it

scarfacered:
350 is wrong, sorry. I just found an error in my logs that says (1410 > 1020)

MWHunter:
Let’s use the TCPShield value then

scarfacered:
Anyone happen to know it?

NoJokeFNA:
Searched for it, haven't found anything.

MWHunter:
They are after the encoder because ProtocolLib.  Maybe paper increased it?

scarfacered:
I made a ticket in their Discord, I'll see what they say.

NoJokeFNA:
And please also comment it **why** you increased it afterwards, so we don't forget it!

scarfacered:
I now believe the root of this issue is using Velocity's player-info-forwarding-mode = "legacy"

After removing the max length on serverAddress locally, the serverAddress variable prints the domain, player IP, player UUID, the name, textures, value, and signature hashes all in one massive string. A total of 1200+ characters. 

There's also seems to be an issue with readString. Bumping it to readString(450) errors later in the method because maxLen isn't multiplied by 4 here:

https://github.com/retrooper/packetevents/blob/2.0/api/src/main/java/com/github/retrooper/packetevents/wrapper/PacketWrapper.java#L375

```java.lang.RuntimeException: The received string length is longer than maximum allowed (1222 > 450)```

MWHunter:
Spigot changed this to Short.MAX_VALUE, so I'm going with spigot here: https://github.com/retrooper/packetevents/commit/db7f8d9cbd5206219c3552007edd20c058a95a04

PR #449, opened 2022-07-04
Improved EquipmentSlot

NoJokeFNA:
I haven't tested it yet with the Wrapper class so it would be great if someone could do it!


retrooper:
Cannot cache the IDs as they change and proxies need the mappings for the version of each client and load it depending on the client. We can't assume its always one ID as it would break on proxies.

NoJokeFNA:
I'll close it for now until we find a overall better solution

PR #452, opened 2022-07-22
Added WrapperPlayServerMapData

LeeGodSRC:
https://wiki.vg/Protocol#Map_Data


api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerMapData.java
```diff
@@ -0,0 +1,177 @@
...
+        if (this.column > 0) {
+            this.writeByte(this.rows);
+            this.writeByte(this.x);
+            this.writeByte(this.z);
+            this.writeByteArray(this.buffer);
+        }
+    }
+
+    @Override
+    public void copy(WrapperPlayServerMapData wrapper) {
```
NoJokeFNA, line 116: No need to call `super.copy` or clone the buffer. Only copy local values

NoJokeFNA (changes requested):

api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerMapData.java
LeeGodSRC: removed

NoJokeFNA: this.buffer = wrapper.buffer.clone(); - this too ^^

LeeGodSRC: done!

api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerMapData.java
```diff
@@ -0,0 +1,175 @@
...
+        this.scale = wrapper.scale;
+        this.icons = new ArrayList<>(wrapper.icons.size());
+        wrapper.icons.forEach(icon -> this.icons.add(icon.clone()));
+        this.column = wrapper.column;
+        this.rows = wrapper.rows;
+        this.x = wrapper.x;
+        this.z = wrapper.z;
+    }
+
+    public static class Icon implements Cloneable {
```
NoJokeFNA, line 127: I'd also suggest moving this to com/github/retrooper/packetevents/protocol/map (or smth. else)

api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerMapData.java
```diff
@@ -0,0 +1,175 @@
...
+            this.type = type;
+            this.x = x;
+            this.y = y;
+            this.rotation = rotation;
+        }
+
+        public Icon(PacketWrapper<WrapperPlayServerMapData> packetWrapper) {
+            short s = packetWrapper.readByte();
+
+            this.type = (byte)(s >> 4 & 15);
```
NoJokeFNA, line 144:
```suggestion
            this.type = (byte) (s >> 4 & 15);
```

api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerMapData.java
```diff
@@ -0,0 +1,175 @@
...
+        public Icon(byte type, byte x, byte y, byte rotation) {
+            this.type = type;
+            this.x = x;
+            this.y = y;
+            this.rotation = rotation;
+        }
+
+        public Icon(PacketWrapper<WrapperPlayServerMapData> packetWrapper) {
+            short s = packetWrapper.readByte();
+
```
NoJokeFNA, line 143:
```suggestion

```

api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerMapData.java
```diff
@@ -0,0 +1,175 @@
...
+    private int column;
+    private int rows;
+    private int x;
+    private int z;
+
+    public WrapperPlayServerMapData(PacketSendEvent event) {
+        super(event);
+    }
+
+    public WrapperPlayServerMapData(int data,
```
NoJokeFNA, line 28:
```suggestion
    public WrapperPlayServerMapData(int data, int scale, boolean locked, @Nullable Collection<Icon> icons, byte[] buffer,
                                    int x, int z, int column, int rows) {
```

NoJokeFNA:
On which versions have you tested it yet?

LeeGodSRC:
I tested it on 1.8 only, for newer version I find information from spigot source and wiki

api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerMapData.java
NoJokeFNA: I actually got confused with the buffer.. my bad. This one has to be copied as you provide it as a parameter. You should probably give it a different name because we are using bytebuf/buffer getters in the PacketWrapper class.

api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerMapData.java
LeeGodSRC: done

api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerMapData.java
LeeGodSRC: done

api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerMapData.java
LeeGodSRC: done

api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerMapData.java
LeeGodSRC: done

NoJokeFNA:
Great! Now 2 last things are still missing:
1. Getters and Setters in the wrapper class for the local parameters
2. I'd suggest adding some abstraction for the MapIcon type. It's currently just a plain byte but it'd to better to have a separate MapType (?) Class which would provide its type as a enun.

api/src/main/java/com/github/retrooper/packetevents/protocol/map/MapIcon.java
```diff
@@ -49,4 +49,51 @@ public int getY() {
...
+        Type(int value) {
+            this.id = (byte) value;
+        }
+
+        public byte getId() {
+            return id;
+        }
+
+        public static Type getById(byte value) {
+            for (Type t : values()) {
```
NoJokeFNA, line 93: Please cache `vales();` -> `private static final Type[] VALUES = values();`

api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerMapData.java
```diff
@@ -113,9 +113,82 @@ public void copy(WrapperPlayServerMapData wrapper) {
...
+
+    public Collection<MapIcon> getIcons() {
+        return icons;
+    }
+
+    public void setIcons(Collection<MapIcon> icons) {
+        this.icons = icons;
+    }
+
+    @Override
```
NoJokeFNA: don't override our buffer, give it some other name. The buffer is already being used in the PacketWrapper class so this could cause issues

retrooper:
Firstly, use serverVersion instead of clientVersion.
Secondly please announce what you're going to work on before hand in either a GitHub issue or in our Discord. Right now I wasn't trying to accept further wrapper pull requests as I am trying to do some refactoring for our wrapper abstraction. 

I'll get to this soon though.

LeeGodSRC:
I made this PR simply because it's missing when I'm trying to use it in one of my project, do I need to create an issue called "Missing wrapper for map data" or something like that?

api/src/main/java/com/github/retrooper/packetevents/protocol/map/MapIcon.java
LeeGodSRC: done

retrooper:
Yea, making an issue beforehand is always good.

markvolkov:
Potential bug on WrapperPlayServerMapData::write line 95?
I think should be this.writeByte((icon.getType().getId() % 27) << 4  | icon.getRotation() & 15);
or just this.writeByte((icon.getType().getId()  << 4)  |  (icon.getRotation() & 15)); assuming getId will only have correct values...
because with current implementation you will get wrong type in MapIcon line 22.
Just need this.type = Type.getById((byte) (s >> 4));   this.rotation = (byte) (s & 15); will still work if you make the above change and get rid of (icon.getType().getId() & 15) << 4 on line 95

markvolkov:
val & 15 is equivalent to val % 16

SquidXTV:
What are we waiting for?

LeeGodSRC:
I'll reopen another PR for this.

PR #468, opened 2022-08-17
An attempt to fix json -> nbt

hinoobers:
Handled LazilyParsedNumber, possibly needs some correction


MWHunter:
Doesn’t look like valid java, or at least java 8

NoJokeFNA:
I'd suggest to remove you change and just use the existing converter..
Before: else if (num instanceof Integer) {
After: else if (num instanceof Integer || num instanceof yourlazystuff) {

PR #477, opened 2022-09-08
Added the ability to set a custom resourceProvider

Souvlakis:
Currently, the retrieval of external resources (json files, such as block mappings) is hard-coded to be through `ClassLoader#getResourceAsStream`.
This PR proposes to add the ability for the end user to change the retrieval algorithm by passing their own `Function<String, InputStream>` to PacketEventsSettings. The function is then called whenever a resource needs to be retrieved.


retrooper:
Curious what you are doing in your project to need to change this.

Souvlakis:
I'm dynamically loading a jar that has packet events shaded in it, thus, standard getResourceAsStream is giving a null output

retrooper:
Appreciate the PR

Issue #341, opened 2022-04-24
Fix joining early with late-bind off on legacy minecraft servers


retrooper:
**Describe the bug**
Joining early with late-bind off on legacy minecraft servers can cause PacketEvents to not notice you joined the server.

**Software brand**
NachoSpigot 1.8.8 (but I assume it works the same on ordinary Spigot)

**Plugins**
ViaVersion 4.2.2

**How To Reproduce**
*Example:*
1. Join with the 1.8.8 vanilla minecraft client while the server starts
2. See error

**Expected behavior**
Be able to join servers that are still booting up without any issues.

MWHunter:
![broken](https://user-images.githubusercontent.com/49797025/173238358-f823f606-9c6a-419c-aa28-6b9f886c2842.png)

Potentially related to this

PR #480, opened 2022-09-25
Forcibly close unhandled connections

Souvlakis:
This is a rather rare one, BUT!
If a plugin using the API takes a while to initialize and handshakes pile up, some connections may not be handled completely, thus, not having the 'splitter' ChannelHandler at the moment PE is initialized.
This PR proposes to drop all such connections.

An example of such behavior would be:
1. A player sends a handshake to the server before PE is initialized
2. Some desync causes Bukkit not to add `splitter` to the channel pipeline before plugins are loaded
3. Once PE loads, a NoSuchElementException is caught in `io.github.retrooper.packetevents.injector.connection.ServerConnectionInitializer`.
The current pipeline handlers in the channel are `[DefaultChannelPipeline$TailContext#0]`

### DIFF
--- a/spigot/src/main/java/io/github/retrooper/packetevents/injector/connection/ServerConnectionInitializer.java
+++ b/spigot/src/main/java/io/github/retrooper/packetevents/injector/connection/ServerConnectionInitializer.java
@@ -87,7 +87,7 @@ public class ServerConnectionInitializer {
              * Taken the above into account, here we just drop all unhandled connections.
              */
             if (channel.pipeline().get("splitter") == null) {
-                channel.unsafe().closeForcibly();
+                channel.close();
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


retrooper:
Perect, were you able to test that this works.

retrooper:
And is there any GitHub reported **Issues** this would fix, let me know so we can close them. If not it is fine too.

Souvlakis:
> Perect, were you able to test that this works.

Yes, had a test plugin pause the main thread in onEnable and tried connecting before/during plugin initialization.
With the patch, the exception no longer occurred.



> And are there any GitHub reported **Issues** this would fix, let me know so we can close them. If not it is fine too.
https://github.com/retrooper/packetevents/issues/341

PR #486, opened 2022-11-06
Safely limit team display name, prefix and suffix to appropriate lengths

vytskalt:
Prevents this:
image

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/util/LegacyFormat.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/util/LegacyFormat.java
@@ -1,0 +1,20 @@
+package com.github.retrooper.packetevents.util;
+
+import net.kyori.adventure.text.serializer.legacy.LegacyComponentSerializer;
+
+public final class LegacyFormat {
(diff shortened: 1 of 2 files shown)


vytskalt:
Can this be reviewed/merged? I'm stuck on a project because of this issue.

Koboo:
tbh, I'd rather throw an exception, if one of the texts should be too long. The substring-handling should be done by the developer himself, otherwise this could lead to unwanted behavior :)

vytskalt:
> tbh, I'd rather throw an exception, if one of the texts should be too long. The substring-handling should be done by the developer himself, otherwise this could lead to unwanted behavior :)

It wouldn't be very efficient to limit a Component into 16 characters yourself. You'd have to convert it into the legacy format, then trim it to 16 characters, then convert it back into a Component and then pass it to the teams packet.

Also, this 16 character limit doesn't exist anymore in modern versions of MC, so if you wanted to take advantage of that while still supporting players on legacy versions you'd have to create two packets (one for 1.12.2 and below players who have the limit and one for 1.13+ players who don't have the limit) for the same thing

PR #509, opened 2022-12-21
Add missing EntityPose

scarfacered:
Error:

```
java.lang.ArrayIndexOutOfBoundsException: Index 14 out of bounds for length 14
	at ac.grim.grimac.shaded.com.github.retrooper.packetevents.protocol.entity.pose.EntityPose.getById(EntityPose.java:51) ~[Grim.jar:?]
```

1.19.3 NMS:

```
package net.minecraft.world.entity;

public enum Pose {
    STANDING,
    FALL_FLYING,
    SLEEPING,
    SWIMMING,
    SPIN_ATTACK,
    CROUCHING,
    LONG_JUMPING,
    DYING,
    CROAKING,
    USING_TONGUE,
    SITTING,
    ROARING,
    SNIFFING,
    EMERGING,
    DIGGING;
}
```

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/protocol/entity/pose/EntityPose.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/protocol/entity/pose/EntityPose.java
@@ -31,6 +31,7 @@ public enum EntityPose {
     DYING,
     CROAKING,
     USING_TONGUE,
+    SITTING,
     ROARING,
     SNIFFING,
     EMERGING,


NoJokeFNA:
Sitting was added in later versions and will therefore throw errors on legacy versions..

NoJokeFNA:
Yes but the problem is that sitting is in between some other enums and not at the end. If you put enums I'm between other enums the ordinal number will logically be changed.

MWHunter:
Closed in favor of https://github.com/retrooper/packetevents/pull/512

PR #529, opened 2023-01-28
Version not being chosen correctly

ammodev:
Fixed that the versions on bungeecord and velocity are not chosen correctly.

### DIFF
--- a/bungeecord/src/main/java/io/github/retrooper/packetevents/bungee/factory/BungeePacketEventsBuilder.java
+++ b/bungeecord/src/main/java/io/github/retrooper/packetevents/bungee/factory/BungeePacketEventsBuilder.java
@@ -99,7 +99,7 @@ public class BungeePacketEventsBuilder {
                         String bungeeVersion =
                                 ProxyServer.getInstance().getVersion();
                         for (final ServerVersion val : ServerVersion.reversedValues()) {
-                        	if(val.getReleaseName().contains(bungeeVersion)) {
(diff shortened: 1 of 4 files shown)


NoJokeFNA:
Don't change packetevents version, 2.0 isn't even official our yet so it's unnecessary

bungeecord/src/main/java/io/github/retrooper/packetevents/bungee/factory/BungeePacketEventsBuilder.java
```diff
@@ -99,7 +99,7 @@ public ServerVersion getVersion() {
                         String bungeeVersion =
                                 ProxyServer.getInstance().getVersion();
                         for (final ServerVersion val : ServerVersion.reversedValues()) {
-                            if (bungeeVersion.contains(val.getReleaseName())) {
+                        	if(val.getReleaseName().contains(bungeeVersion)) {
```
NoJokeFNA, line 102:
```suggestion
                        	if (val.getReleaseName().contains(bungeeVersion)) {
```

NoJokeFNA (changes requested):

bungeecord/src/main/java/io/github/retrooper/packetevents/bungee/factory/BungeePacketEventsBuilder.java
ammodev: Not sure why VSCode sorted the imports and applied Prettier to everything but this one if. Resolved.

PR #540, opened 2023-02-18
feat: Add option in settings to disable color downsampling + fix actions/upgrade gradle

cheaterSK:
This PR adds an option for developers to disable forced color downsampling, which essentially allows for making custom handlers on pre-1.16 servers. This changes nothing by default, only adds an option for anyone that may need it.


cheaterSK:
I upgraded gradle to 8.0.1 and (hopefully) fixed the issues that came with it. I also changed actions to use gradle wrapper instead of gradle to avoid issues like this in future.

PR #546, opened 2023-03-08
Expose NPC channels and remove unnecessary parameters

jtJava:
I'm exposing the netty channels to make it easier for people to have their own implementations of the NPC class. In my case, I need to retrieve the viewers in order to remove the nametag visibility of the NPC and spawn a "Hologram" to remove the below name scoreboard objective from the NPC.


MachineBreaker:
This is actually a breaking change, can't be improved?

jtJava:
I mean, it has to be done at some point. There's no reason for that parameter to be there.

PR #551, opened 2023-03-14
Fix getEnchantmentLevel NPE

MachineBreaker:
The method didn't had any NPE protection, this caused some issues like in GrimAC


MachineBreaker:
No idea if the fix should use continue or returning 0, what do you guys think?

SamB440:
There seems to be another problem similar to this, in `getEnchantmentTypeFromTag`. It seems that sometimes items may not have the `minecraft:` prefix in the id field, which then of course breaks the `return EnchantmentTypes.getByName(id);`. I'm not sure in what cases this happens, I just randomly discovered it when using a knockback sword one day (which would also cause falses with Grim).

PR #557, opened 2023-03-17
Use Pooled ByteBuffer in SpigotReflectionUtil#decodeBukkitItemStack

MachineBreaker:
Using a pooled ByteBuf is more efficient because it avoids the overhead of allocating and deallocating ByteBufs repeatedly. It also reduces memory fragmentation and improves garbage collection efficiency.


MachineBreaker:
GitHub just killed me

PR #560, opened 2023-03-18
Fix 1.19.4 entity mappings

SamB440:
This fixes the following:

- 1.19.4 version not registered in EntityTypes
- Typo in mappings, `text-display` -> `text_display`
- The entity id mappings were simply wrong, Mojang changed the order of a bunch of entities around (I think it's alphabetical order now?)

I checked the 1.19.3 -> 1.19.4 diff, verified the ids with a client-side mod, and tested this with a plugin, before it would spawn a strider when I wanted text_display, with this change it is correct.


MWHunter:
Didn't see this PR before I fixed the mappings myself.

Never using ChatGPT to update mappings again...

PR #564, opened 2023-03-26
Fix WrapperPlayClientChatCommand and WrapperPlayClientChatMessage

cheaterSK:
This PR fixes some weird bug which I am not sure what caused it, basically was just erroring when you tried to instantiate it. In addition it also adds read/write for signed command arguments. Tested on 1.19-1.19.4.


retrooper:
Thanks for fixing this issue!

PR #575, opened 2023-04-15
Added a Conversion Util to convert to/from PacketEvents to Bukkit Particles

Aseeef:
Tested on 1.16 and 1.12 but should work on other versions too.


Aseeef:
By the way, I removed that  `import org.bukkit.craftbukkit.v1_16_R1.CraftParticle;` import when I updated my branch from remote.

Aseeef:
@retrooper

PR #576, opened 2023-04-20
Added Spoof (https://spoof.abyssdev.net) support.

sysdm-del:
This pull request adds support for the popular Spoof plugin, where packets are not injected if a players' netty channel is instanceof a fake channel (SpoofedChannel).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -49,6 +49,11 @@ repositories {
     maven {
         url = uri('https://repo.maven.apache.org/maven2/')
     }
+
+    maven {
+        url = uri('https://maven.abyssdev.net/api')
+    }
+
     mavenCentral()
 }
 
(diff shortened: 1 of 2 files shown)


sysdm-del:
> Why does packetevent has to add support for a plugin and not the plugin has to add support for pe?

Because it doesn't work? It's an issue with PacketEvents' injection, they either have to add this or some bypass of some sort (silencing for example).

NoJokeFNA:
Change it back to default Code style. If you are just changing some lines there is no need to completely change it for one single method. Removed the empty lines that you have added and everything else you have changed regarding to it.

sysdm-del:
> Change it back to default Code style. If you are just changing some lines there is no need to completely change it for one single method. Removed the empty lines that you have added and everything else you have changed regarding to it.
Fixed

MachineBreaker:
Why the 'Spoof' plugin can't just cancel the inject event? There's a reason for the event to exist in the first place

AbhigyaKrishna:
I still don't get the reason behind doing this. A library plugin doesnt have to make compatibility changes for every single plugin it encounters. It defeats the purpose of a library. Instead the "Spoof plugin" should make the changes and cancel the injection of their spoof channel. A library should load before every other plugin. And if you are shadowing packetevents you should be accountable for its compatibility with other non library plugins.

sysdm-del:
> Why the 'Spoof' plugin can't just cancel the inject event? There's a reason for the event to exist in the first place

Ah my bad, never saw this.

Closing this PR.

PR #577, opened 2023-04-23
Make entity mappings unload after types load

Pyrbu:
Currently the static block that unloads the entity id mappings in the EntityTypes class is ran before any of the types are defined, this moves it down to the end of the file and adds a comment telling others not to move it.

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/protocol/entity/type/EntityTypes.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/protocol/entity/type/EntityTypes.java
@@ -54,12 +54,6 @@ public class EntityTypes {
             ClientVersion.V_1_12,
             ClientVersion.V_1_13);
 
-
-    static {
-        TYPES_BUILDER.unloadFileMappings();
-        LEGACY_TYPES_BUILDER.unloadFileMappings();
-    }
-
     public static EntityType define(String key, @Nullable EntityType parent) {
         TypesBuilderData data = TYPES_BUILDER.define(key);
         TypesBuilderData legacyData = LEGACY_TYPES_BUILDER.define(key);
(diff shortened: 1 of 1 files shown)


Jaren8r:
The static block is always the first thing that is executed in a class. The location of it shouldn't matter.

Pyrbu:
> The static block is always the first thing that is executed in a class. The location of it shouldn't matter.

The static variable initialization gets compiled down to just another static block so they get executed in order

Pyrbu:
Current decompiled code:
image

api/src/main/java/com/github/retrooper/packetevents/protocol/entity/type/EntityTypes.java
```diff
@@ -308,4 +302,10 @@ public static EntityType getByLegacyId(ClientVersion version, int id) {
     public static final EntityType TEXT_DISPLAY = define("text_display", DISPLAY);
     public static final EntityType INTERACTION = define("interaction", DISPLAY);
     public static final EntityType SNIFFER = define("sniffer", ABSTRACT_ANIMAL);
+
+    // This HAS to be at the end of the file so the types load properly, please do not move it
```
Jaren8r, line 306:
```suggestion
    // This HAS to be at the end of the file so the types unload properly, please do not move it
```

Jaren8r (approved):
🤦 I didn't realize that the define blocks were causing it to load right after being unloaded and assumed that a much more catastrophic error would be happening if it was out of order (this is what I get for looking at things tired). Looks good, but I'd clarify the comment to say unload instead of load.

Pyrbu:
It would’ve been much more catastrophic but in the define method there is a check that just re-loads it if its unloaded so its more of a wasted memory issue

Jaren8r:
Yep, I noticed that when I dug deeper.

AbhigyaKrishna:
The issue in this pr is resolved by #648

PR #583, opened 2023-05-16
Fix folia support

mani1232:


powercasgamer (changes requested):
The `RegionizedServer` class exists..?

mani1232:
Yes

powercasgamer:
> Yes

Then there's no point in making this PR since the RegionizedServer class already exists.

mani1232:
RegionizedServer not exists*

powercasgamer:
> RegionizedServer not exists*

It does, 
image

mani1232:
I checked, when the server is started, the plugin does not find it, I changed it everywhere in my plugins.

powercasgamer:
> I checked, when the server is started, the plugin does not find it, I changed it everywhere in my plugins.

Can't reproduce: 
image

mani1232:
hm...

PR #591, opened 2023-06-12
Fix missing block tag states

SamB440:
It seems the commit https://github.com/retrooper/packetevents/commit/9975706ba886c1cb9cf447b0fdae5ad3ced80d83 accidentally got rid of bamboo.

I've also added the new 1.20 blocks to the tags, such as cherry.

I checked the source and diff (1.19.4-1.20), I don't think I missed anything, but I probably did.

Also, `SUSPICIOUS_GRAVEL`, `PITCHER_CROP` AND `PITCHER_PLANT` seem to be missing as states, unsure if that is intentional.


retrooper:
What makes you think those states are missing?

retrooper:
Please double check with latest commit? Cause your PR is only one hour old.

SamB440:
I am an idiot and didn't update my fork, fixed now

retrooper:
Is the REPLACEABLE_BY_TREES field named that intentionally, when the value is REPLACABLE_PLANTS?

SamB440:
> Is the REPLACEABLE_BY_TREES field named that intentionally, when the value is REPLACABLE_PLANTS?

It got renamed... unsure if such things should be cared about considering it breaks API (I forgot to update the string anyway).

I'll change it back since that's probably best

retrooper:
I don't know, if the value got changed aswell, then a new field would have to be made? I'm not sure about that. But thank you for the PR, is it ready to be merged?

SamB440:
Added stone buttons so we're the same as vanilla, and fixed bamboo_blocks. Should be good now.

retrooper:
Thank you for your PR!

PR #593, opened 2023-06-14
Add 1.20.1 server version

booky10:
The client version for 1.20.1 doesn't need to be created, as they to my understanding only reflect protocol versions.


retrooper:
Missed the PR, but thanks though.

PR #594, opened 2023-06-17
1.19.3 ChatMessage fix

AbhigyaKrishna:
retrooper asked for this


retrooper:
You can only do this if the chatComponent field is empty, you can't just always do it. That defeats the purpose of that field in the protocol.

retrooper:
Thanks for identifying the issue though.

AbhigyaKrishna:
There is no chat component field in 1.19.3+. The body comes only with a plain text, no json message.
image

retrooper:
There literally is, look at Unsigned Content, it is optional. If it is not present, then we must use the plain text...

retrooper:
You can also compare with source.

AbhigyaKrishna:
Unsigned content is also there in 1.19.1 chat message packet, along with the chat content component. I am unsure about the unsigned content being what is needed here.
image
image

retrooper:
Let me look into 1.19.1

retrooper:
Ok, perfect, 1.19.3+ chat components work.

retrooper:
You undid my 1.19.3+ changes and worked on 1.19.1...

AbhigyaKrishna:
The chat message would be processed this way so yes

retrooper:
Why did you revert my 1.19.3 fixes?

retrooper:
It's not correct.

retrooper:
Why not make 1.19.1 and 1.19.3 code consistent?

AbhigyaKrishna:
Two reason:
1. You changed the unsigned chat component field which is nullable to a not null value with the plain text.
2. The chat message would be processed this way. I have confirmed this by decompiling 1.19.1, 1.19.3 and 1.19.4.

PR #596, opened 2023-06-27
More block tag fixes

SamB440:
Round 2...

- Fixed BAMBOO_MOSAIC_STAIRS not being in STAIRS block tag
- Reordered slabs to be in parity with vanilla
- Made step block sound tags same as vanilla
- Added SWORD_EFFICIENT and REPLACEABLE from 1.20

Also added remaining new 1.20 tags... even though some are useless because they contain one block, but already added the sniffer one so might as well do the rest & for parity.

Will work on ItemTags in another PR later.


retrooper:
Thank you

PR #597, opened 2023-06-27
Update ItemTags to 1.20

SamB440:
Updates ItemTags to 1.20, copied directly from source then manually fixed... can't be any mistakes in this one.


retrooper:
Thank you for your PR.

PR #598, opened 2023-07-04
Fix chest block attribute order

clrxbl:
@AoElite


retrooper:
Thank you for the PR fixing the issue, but this is not a long-term solution. Since our mappings are based on https://github.com/ViaVersion/Mappings/blob/main/mappings/mapping-1.20.json we would have to manually do this each time? Any ideas? We could make an issue on their repository to inform them?

AoElite:
Kenny doesn't want to fix the mappings; they didn't really elaborate as to why. So, we should probably just sort the attributes ourselves in the code.

retrooper:
Yes, unfortunate.

PR #599, opened 2023-07-08
Limit objective display name to 32 characters

vytskalt:
same as #486 but for objective display name


retrooper:
PR was not merged due to conflicts, but latest commit fixes this.

PR #601, opened 2023-08-05
Updater updates

mpc19:
I think at the moment, the updater is designed in a way which isn't entirely useful for a user. Currently, it will try for a total of 160 seconds before giving up, producing stack traces each time, which is not great for readability.
**_Before_**
before
_**After**_
after

Also changed some uses of System.out.println to the log manager, so people know the message is from packetevents


retrooper:
Will look into this. Have you tested this to work?

PR #607, opened 2023-09-12
1.20.2 update

booky10:
~~Still needs testing after 1.20.2 actually releases~~
Tested on paper 1.8.8, 1.20.1 and 1.20.2 with GrimAC


retrooper:
The PR will be merged, a stable release cannot be put out yet because there are still issues on proxies. (specific to 1.20.2)

retrooper:
Thanks for the PR

PR #610, opened 2023-10-01
Retain backward compatibility with adventure 5.0.0

AbhigyaKrishna:
Added compatibility for adventure 5.0.0


MachineBreaker:
Won't this make a performance hit every time if we go through the exception?

AbhigyaKrishna:
Then don't use ancient adventure version.

PR #618, opened 2023-11-03
Expose block id to BaseChunk

SoSeDiK:
Exposes getting block state id in BaseChunk.

I wasn't sure on how to name the method (the codebase uses at least `blockId`, `globalId` and `combinedId`), so since it's (chunk data) usually used alongside other block packets, I went with how it's called in them ([WrapperPlayServerBlockChange](https://github.com/retrooper/packetevents/blob/363931875fe619711ddc2646a596355a30648246/api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerBlockChange.java#L83) and [WrapperPlayServerMultiBlockChange](https://github.com/retrooper/packetevents/blob/363931875fe619711ddc2646a596355a30648246/api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerMultiBlockChange.java#L177)), but it can be renamed upon request.


retrooper:
Thanks for the PR!

Issue #621, opened 2023-11-14
Relocate breaks kyori serializer


rexlManu:
The relocation at https://github.com/retrooper/packetevents/blob/2.0/spigot/build.gradle#L11 also relocates the base ComponentSerializer Class, which breaks other plugins which depend on kyori and add as example legacy component serializer.

The exception that will be thrown is a java.lang.NoClassDefFoundError: net/kyori/adventure/text/serializer/ComponentSerializer.

When trying to shade kyori api with it, packetevents throw a LinkageError (Class Loader Conflict: The error indicates a class loader conflict involving net.kyori.adventure.text.Component)

rexlManu:
Fixed with my pr

PR #622, opened 2023-11-15
Relocate only the gson and legacy

rexlManu:
This pull request addresses issue #621, which arises from the net.kyori.adventure.text.serializer being part of the API. However, only net.kyori.adventure.text.serializer.gson and net.kyori.adventure.text.serializer.legacy are additional packages for text serialization.

The core issue occurs when a plugin relies on PacketEvents and utilizes Kyori Adventure (for example creating a packet wrapper that uses it). It needs to include and distribute a Text Serializer (like legacy). During runtime, an error is thrown due to the non-discovery of the base interface from the API, as it's relocated.

Notably, this problem is specific to Spigot/BungeeCord environments and does not occur on Paper servers.

### DIFF
--- a/bungeecord/build.gradle
+++ b/bungeecord/build.gradle
@@ -6,7 +6,8 @@ dependencies {
 }
 
 shadowJar {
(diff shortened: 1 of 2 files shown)


retrooper:
Thanks

PR #623, opened 2023-11-17
1.20.3/1.20.4 update

booky10:
~~Ofc still very much WIP, 1.20.3 ~~doesn't even have pre-releases yet~~ isn't released yet.~~

There is a major change to the protocol, components are now serialized as NBT instead of JSON.

(https://gist.github.com/booky10/e2a3682b114900f70c89c0dc12da483e has some code I've used for extracting data from vanilla)


booky10:
I've tested most of the new stuff, works on spigot 1.20.3.

AoElite:
@booky10 Have you tested it on older server versions as well?

retrooper:
in the new NBT-Serializer for Components there's this one bug on 1.20.3/1.20.4 https://pastebin.com/mzM8TtRg

retrooper:
Thank you for the PR!

PR #627, opened 2023-12-16
feat: add equals method to StateType

micartey:
Adding the `equals` method to the StateType class to easily distinguish and compare in a Map / List.


retrooper:
Why do you return false if the classes are equal?

AoElite:
7ff8c404029ada0e912c7cb70449d2030de10567

PR #661, opened 2024-01-21
Read and write entity eye height in Vibration particle wrapper

joshuaprince:
Fixes https://github.com/joshuaprince/CoordinateOffset/issues/9

Vibration particle data for vibrations emitted by entities contains an "entity eye height" field. The ParticleVibrationData wrapper for WrapperPlayServerParticle did not read or write this field, so any usage of this wrapper would corrupt the packet and cause players to be kicked when they moved near an entity that can receive vibrations (i.e. Wardens).

Tested on 1.19.4 and 1.20.4. This field has been present and unchanged since 1.19's changes to the Vibration particle (which is also the first fully-released version to add Wardens):
https://wiki.vg/index.php?title=Data_types&oldid=17608#Particle


booky10:
This should check for the server version as the field was added in 22w12a (1.19 snapshot). The particle type already existed in 1.18.2.

joshuaprince:
> This should check for the server version as the field was added in 22w12a (1.19 snapshot). The particle type already existed in 1.18.2.

Updated to check for 1.19+ to read this field.

api/src/main/java/com/github/retrooper/packetevents/protocol/particle/data/ParticleVibrationData.java
```diff
@@ -174,6 +198,10 @@ public static void write(PacketWrapper<?> wrapper, ParticleVibrationData data) {
             wrapper.writeBlockPosition(data.getBlockPosition().get());
         } else if (data.getType() == PositionType.ENTITY) {
             wrapper.writeVarInt(data.getEntityId().get());
+            if (wrapper.getServerVersion().isNewerThanOrEquals(ServerVersion.V_1_19)
+                    && data.getEntityEyeHeight().isPresent()) {
+                wrapper.writeFloat(data.getEntityEyeHeight().get());
+            }
```
booky10, line 204: IMO this should write a default value, even if no entity eye height has been specified. I would prefer having some strange values, instead of having an exception kick players from the server.

booky10 (approved):
:+1:

PR #668, opened 2024-02-02
Request signatures from Mojang API

honkling:
The Mojang API doesn't provide signatures by default, so it's impossible currently to get the signature of a texture using the Mojang API utility.

This pull request resolves this.


retrooper:
Thanks for the PR.

PR #670, opened 2024-02-03
(WIP) Sponge support

SamB440:
Adds Sponge support. A lot is copied from the Spigot module as they are fundamentally similar and via runs on Sponge too.

The only conversion util I made is itemstack conversion. Most things can be done via Sponge API or VanillaGradle anyway.

~~Also fixed use of deprecated `java.sourceCompatibility` and `java.targetCompatibility` (need to confirm it compiles to 8 bytecode again).~~


settings.gradle
```diff
@@ -1,8 +1,15 @@
+pluginManagement {
```
NoJokeFNA, line 1: Why are you adding this into settings.gradle?

build.gradle
```diff
@@ -26,8 +26,10 @@ subprojects {
                                        "net.kyori:adventure-text-serializer-legacy:${adventureVersion}",
                                        "net.kyori:adventure-nbt:${adventureVersion}"]
 
-    java.sourceCompatibility = JavaVersion.VERSION_17
-    java.targetCompatibility = JavaVersion.VERSION_1_8
+    java {
```
NoJokeFNA, line 29: I would prefer to have Gradle changes in separate PRs

settings.gradle
SamB440: I was messing with vanilla gradle at one point, gone now.

SamB440:
Update for those unaware: viaversion support doesn't seem to be possible because on Sponge they replace the vanilla handler, instead of adding their own.

SamB440:
I think I might migrate the current injection stuff to be mixin-based like fabric

retrooper:
> Update for those unaware: viaversion support doesn't seem to be possible because on Sponge they replace the vanilla handler, instead of adding their own.

They have fixed this recently.

retrooper:
You can mark this as a PR.

retrooper:
Is this fully functional?

SamB440:
Yes

retrooper:
Thank you a lot for porting PacketEvents to Sponge!

retrooper:
Next update, it should be listed as supported.

Issue #671, opened 2024-02-06
Missing Play.Server.ACTION_BAR Wrapper


gabber235:
**Describe the bug**
Using a `PacketListenerAbstract` I want to listen for all the actionbars send to a player. As far as I am aware there are two ways for the server to send an action bar. The first is using `Play.Server.SYSTEM_CHAT_MESSAGE` and the second is using `Play.Server.ACTION_BAR`. With the systems chat there is no problem. However for the `Play.Server.ACTION_BAR` I cant seem to find the correct Wrapper for it. I thought it was the `WrapperPlayServerTitle` as it is the only wrapper that even mentions the action bar. But it throws the following error: 

```
java.lang.ArrayIndexOutOfBoundsException: Index 10 out of bounds for length 6
    at com.github.retrooper.packetevents.wrapper.play.server.WrapperPlayServerTitle$TitleAction.fromId(WrapperPlayServerTitle.java:236) ~[packetevents-spigot-2.2.0.jar:?]
    at com.github.retrooper.packetevents.wrapper.play.server.WrapperPlayServerTitle.read(WrapperPlayServerTitle.java:76) ~[packetevents-spigot-2.2.0.jar:?]
    at com.github.retrooper.packetevents.wrapper.PacketWrapper.readEvent(PacketWrapper.java:233) ~[packetevents-spigot-2.2.0.jar:?]
    at com.github.retrooper.packetevents.wrapper.PacketWrapper.<init>(PacketWrapper.java:143) ~[packetevents-spigot-2.2.0.jar:?]
    at com.github.retrooper.packetevents.wrapper.PacketWrapper.<init>(PacketWrapper.java:133) ~[packetevents-spigot-2.2.0.jar:?]
    at com.github.retrooper.packetevents.wrapper.play.server.WrapperPlayServerTitle.<init>(WrapperPlayServerTitle.java:49) ~[packetevents-spigot-2.2.0.jar:?]
```

I want to be able to read the new https://wiki.vg/Protocol#Set_Action_Bar_Text packet. How can I do this?

I already asked this question, and the consensus seemed to be that there was no Wrapper made for this packet yet.

**Software brand**
Paper 1.20.4

**Plugins**
PacketEvents, [Typewriter](https://modrinth.com/plugin/typewriter) (my plugin)

**How To Reproduce**
```kotlin
class MyHandler : PacketListenerAbstract(PacketListenerPriority.HIGH) {

    fun initialize() {
[...]

PR #674, opened 2024-02-07
Added WrapperPlayServerSetActionBarText

gabber235:
Create a wrapper for the [Set Action Bar Text Packet](https://wiki.vg/Protocol#Set_Action_Bar_Text). Fixing #671.

It is mostly a copy of [WrapperPlayServerSetTitleText](https://github.com/retrooper/packetevents/blob/2.0/api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerTitle.java) as the packets have the same fields.


booky10:
Is there a reason for copying the deprecated fields/methods?

PR #691, opened 2024-03-02
Fix re-encode for new bytebufs that were bigger than the received one

ytnoos:
If you tried to mark for re-encode a packet containing ItemStack and the received ByteBuf didn't have enough capacity to write the new data an error occurred. Allocating a new ByteBuf will allow us to always write the required bytes


api/src/main/java/com/github/retrooper/packetevents/util/PacketEventsImplHelper.java
```diff
@@ -103,7 +107,7 @@ public static PacketReceiveEvent handleServerBoundPacket(Object channel, User us
                 task.run();
             }
         }
-        return packetReceiveEvent;
+        return buffer;
```
retrooper, line 110: Why is the buffer returned, and the method return changed?

ytnoos: Because we have to retain the correct buffer, previously the buffer needed to be retained was always the `input` one, now if the packet is marked for `re-encode`, we clear the `input `buffer and allocate a new one where to write data.

retrooper:
Although I feel like there might be some method to fix this and reset the capacity, I'll just merge it.

PR #694, opened 2024-03-04
Correct NBT component style serialization

Kamilkime:
During development of one of our plugins we've noticed that display name components in the PLAYER_INFO_UPDATE packets were not colored correctly. Turns out it's a small error in the component style NBT serialization - the code expected the style to be in the `style` compound (I think adventure holds it that way in its components, maybe that's why it was written that way), when in fact the styling elements are directly in the component compound.

You can see the NBT styling description [here](https://wiki.vg/Text_formatting#Styling_fields) - as you can see it's e.g. just a `color` tag, it's not inside any other styling tag.

This PR just simply gets rid of the additional `style` compound, the styling data is read directly from the main component compound, and is written directly into the component compound.

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/util/adventure/AdventureNBTSerializer.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/util/adventure/AdventureNBTSerializer.java
@@ -73,6 +73,7 @@ public class AdventureNBTSerializer implements ComponentSerializer<Component, Co
                 throw new IllegalStateException("Don't know how to deserialize " + nbt.getType() + " to text");
             }
         };
+
         String text = reader.read("text", textFunction);
         if (text == null) text = reader.read("", textFunction);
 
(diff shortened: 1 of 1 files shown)


AbhigyaKrishna:
Are you sure that style tags are serialized in the root tag and not in a separate "style" parent tag?

SamB440:
I tested this pr after updating and realising all messages I was sending were white - however there still seems to be a problem as the messages I send are now forced to their closest vanilla named colour.

Also everything seems to work fine in 1.20.2 even without this pr... but breaks in 1.20.4

Kamilkime:
> Are you sure that style tags are serialized in the root tag and not in a separate "style" parent tag for all versions?
> Also everything seems to work fine in 1.20.2 even without this pr... but breaks in 1.20.4

I think these two questions are actually similar - as you can see [here](https://wiki.vg/Text_formatting#Text_components), the NBT format for these messages was introduced in 1.20.3, so that would be why earlier versions are ok, iirc there is a check somewhere in packetevents which uses the NBT serializer only for the 1.20.3+. As to am I 100% sure if it all works always... Maybe not 100%, but our plugin is being tested on 1.16.5, 1.19.4, and 1.20.4, and it seems that everything works ok on all these versions, no errors to be seen.

> however there still seems to be a problem as the messages I send are now forced to their closest vanilla named colour

Right, I didn't notice it, it would be a problem in our plugin too 😅 

But fortunately it's not a bug, maybe just a lacking documentation - there is a color downsampling setting, so that they look ok on pre-1.16 clients. However this serializer is really only 1.20.3+, so maybe this setting is a bit unnecessary here. In any case - you can use `PacketEventsSettings#downsampleColors(false)` to turn that off during the packetevents load. Now it looks great, all colors are definitely RGB, thanks for noticing that 😃

image

Kamilkime:
So... Are we doing anything with this PR? It's quite small and quick :/

AbhigyaKrishna:
Thank you for the pr

PR #699, opened 2024-03-13
Added CommonPacketWrapper

Rubenicos:
This pull request is intended to make the code similar to Mojang code

I tried my best to avoid any API break, but some interactions require to change them completely (see breaking changes below)

### Additions

* Method to get `ConnectionState` from `PacketType`
* `CommonPacketWrapper` class, same has `PacketWrapper` but stores a `ConnectionState` to wrap different packets with the same structure but different protocol state
* Any wrapper previously covered by PacketEvents code as its common type (for example `WrapperClientPong` to replace `WrapperConfigClientPong` and `WrapperPlayClientPong`)
* Fallback ordinal value for `WrapperClientResourcePackStatus.Result`

### Changes

* All the wrappers replaced by its common wrapper type, now are marked as `Deprecated`

### Breaking changes

* `WrapperConfigClientResourcePackStatus` and `WrapperPlayClientResourcePackStatus` no longer have `Result` enum, use `WrapperClientResourcePackStatus.Result` instead

### Notes

* `WrapperServerDisconnect` can also replace `WrapperLoginServerDisconnect` but Mojang still maintain two separated classes for those packets
* `WrapperServerPing` can also create a wrapper for `WrapperPlayServerPing` on `CONFIGURATION` state

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/protocol/packettype/PacketType.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/protocol/packettype/PacketType.java
@@ -201,6 +201,11 @@ public final class PacketType {
             public PacketSide getSide() {
                 return PacketSide.CLIENT;
             }
+
+            @Override
+            public ConnectionState getState() {
+                return ConnectionState.HANDSHAKING;
+            }
         }
 
         public enum Server implements PacketTypeConstant, ClientBoundPacket {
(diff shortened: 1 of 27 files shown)


retrooper:
I see why you implemented it like that, but I cannot accept all the breaking changes.

PR #718, opened 2024-04-23
Draft implement proper adventure API support to packet events

Tofaa2:
This PR includes:

User class now inherits and *mostly* (Missing  sound sending,  which is still a WIP) implements Audience functionality.

introduces a GroupedAudience interface that allows for people to filter members easily and send grouped packets and content to the client

Introduces WrapperPlayServerBossBar

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/protocol/world/BlockFace.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/protocol/world/BlockFace.java
@@ -153,15 +153,16 @@ public enum BlockFace {
         }
     }
 
-    /** Returns int representing cardinal directions with south as the starting index at 0. UP and DOWN return -1
+    /**
+     * Returns an integer representing cardinal directions with south as the starting index at 0. UP and DOWN return -1
      * <p>
      * Used by Mojang for calculations involving mob AI, decorations, paintings, item frames, maps,
      * structure generation, campfires, stairs, connecting blocks, and flowerbeds.
-     * Also used for getting Directions and Blockfaces as Rotations.
+     * Also used for getting directions and block faces as rotations.
      *
-     * @return int horizontal id
+     * @return integer horizontal id
      */
-    public int getHorizontalID() {
+    public int getHorizontalId() {
         switch (this) {
             case DOWN:
             case UP:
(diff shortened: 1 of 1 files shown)


api/src/main/java/com/github/retrooper/packetevents/protocol/player/User.java
```diff
@@ -84,6 +104,151 @@ public void setConnectionState(ConnectionState connectionState) {
...
+            stay = (int) (times.stay().toMillis() / 20);
+            out = (int) (times.fadeOut().toMillis() / 20);
+            in = (int) (times.fadeIn().toMillis() / 20);
+        }
+        else {
+            stay = 20;
+            out = 20;
+            in = 20;
+        }
+        sendTitle(title, subtitle, in, stay, out);
```
booky10, line 136: are you sure this is what sendTitlePart is supposed to do? wont this override the title, if you send a title-part and then a subtitle-part?

Tofaa2: > are you sure this is what sendTitlePart is supposed to do? wont this override the title, if you send a title-part and then a subtitle-part?

It would override yeah

api/src/main/java/com/github/retrooper/packetevents/protocol/player/User.java
```diff
@@ -84,6 +104,187 @@ public void setConnectionState(ConnectionState connectionState) {
...
+        else {
+            stay = 20;
+            out = 20;
+            in = 20;
+        }
+
+        ServerVersion version = PacketEvents.getAPI().getInjector().isProxy() ? getClientVersion().toServerVersion() :
+                PacketEvents.getAPI().getServerManager().getVersion();
+        boolean modern = version.isNewerThanOrEquals(ServerVersion.V_1_17);
+        PacketWrapper<?> animation;
```
booky10, line 140: Shouldnt this also default to null if there wont be any TIMES-part sent?

AbhigyaKrishna:
Is these really necessary? Why do we need to implement a adventure platform in a packet library?
A separate addon would be more appreciable rather than bloating this library.

Issue #726, opened 2024-05-02
Dropping support for "stupid" versions


ThrowerWay:
Would you be against dropping support for stupid versions of the game? ~
i.e. any version that is not 1.8.8, 1.12.2, 1.20.x?
Would reduce bloat a lot and those 3 versions make up 99.9% of non dead servers.

Even then, its kinda a stretch to call 1.12 relevant in 2024, 1.12 was for 2b2t and 2b2t clones, but since 2b2t has updated 1.12 is not really relevant.

Anthony01M:
Some gamemodes are kept intentionally on old versions in some servers (bedwars is a primary example);
Plus you are forgetting that a lot of the plugins would like to keep support for old versions and many would switch from PacketEvents if old versions support was dropped due to this.

OakLoaf:
> Would reduce bloat a lot and those 3 versions make up 99.9% of non dead servers.

Where are you getting this "99.9%" statistic from? I personally run 1.20 but as Anthony said it would definitely turn people away from PacketEvents as a library if that many versions were dropped

retrooper:
I understand your point, but before actively going to make this change in a PR, wait for an official response. That way no time is wasted. You can certainly lead your own fork if you deem this change as necessary, Thank you for the suggestion, but I do not support the idea and would therefore have to close the PR.

PR #729, opened 2024-05-02
Remove legacy code and update to java 17

ThrowerWay:
Removes code for legacy versions, improving performance, maintainability and debloating the codebase.


Bram1903:
It doesn't fix `#726`, but rather closes it.

Nevertheless, I vote against this PR, as I believe that when we stop supporting all these versions, many developers might switch to ProtocollLib because they are either forced to support those versions or want their plugin to be as accessible as possible.

Bram1903:
> > Also switching to Java 17 might create issues on certain 1.8 forks.
> 
> This PR drops support for all versions except 1.20.5

I hope this is some really bad joke right?

PR #767, opened 2024-05-11
fixed entity id assign bug

ItzZisker:
As what I saw in NMS 1_8_R3, the entityCount first assigned then gets incremented, but in newer versions they migrated to AtomicInteger values in which they used the incrementAndGet() method which increments then gets the value. I discovered this when weird stuff starts happening, for example: NPC spawned but froze the player, in which I suspected about the entity id being the same on both NPC & player.
NMS code (entity constructor):
``` java
    public Entity(World world) {
        this.id = entityCount++;
[...]
````


retrooper:
Thanks for the PR.

PR #790, opened 2024-05-19
Gradle build chain rework with kotlin DSL

AbhigyaKrishna:
Rewritten whole of the build chain in kotlin dsl with best practices.

Tasks:
- [x] Rewrite with kotlin dsl
- [x] Version catalog for dependencies
- [x] Use gradle convention plugin instead of subproject configuration
- [x] Improve build chain
- [x] Rewrite mapping compression task in kotlin


SamB440:
Perhaps consider updating the JDK to 21 as that's what Minecraft uses now. Fabric will need it when updating, I already had to do this for my Sponge PR.

Bram1903:
Great PR! Certainly a huge improvement! Versioning is now much easier to maintain 💪🏻

buildSrc/src/main/kotlin/packetevents.library-conventions.gradle.kts
```diff
@@ -0,0 +1,73 @@
...
+}
+
+dependencies {
+    compileOnly("org.jetbrains:annotations:23.0.0")
+}
+
+java {
+    withSourcesJar()
+    disableAutoTargetJvm()
+}
```
Bram1903, line 23: Perhaps, as @SamB440 said earlier in this PR, we should set the toolchain to use Java 21. Due to your compiler settings, this will still compile back to Java 8.

```suggestion
java {
    toolchain.languageVersion = JavaLanguageVersion.of(21)
    withSourcesJar()
    disableAutoTargetJvm()
}
```

AbhigyaKrishna: We should be forcing the language version of the code to be Java 8. Hence, this shouldn't be done. As of @SamB440, he is introducing the sponge module which would require java version to be 21. This shouldn't be set globally as our own code should remain Java 8. And for the fabric, its already set to Java 17.

Bram1903: Thanks for the clarification! About Fabric specifically, that should probably be updated from Java 17 -> 21.

retrooper: Is it necessary and is Fabric backwards compatible?

retrooper: Not that backwards compatibility is essential on any platform other than Spigot, but if it is not necessary why bother now?

AbhigyaKrishna: This pr is not for updating fabric, I shouldn't merge two key update in one pr.

Bram1903: I agree with you Abhigya. Fabric isn't relevant for this PR. Anyway, the PR is really nice! It's much easier to navigate through the build chain now.

retrooper: 👍

Issue #793, opened 2024-05-21
doesn't work on paper 1.20.6 without remapping


IssueDescription:
**Describe the bug**
If you have remapping disabled on paper, packetevents doesnt work, because packet events is still using spigot obfuscated nms calls, there should be support for mojang mapped, which is what paper is using now since 1.20.5

```
[14:32:21 INFO]: [packetevents] Enabling packetevents v2.3.1
[14:32:21 WARN]: java.lang.NoSuchMethodException: net.minecraft.world.item.ItemStack.<init>(net.minecraft.world.item.Item,int)
[14:32:21 WARN]:        at java.base/java.lang.Class.getConstructor0(Class.java:3761)
[14:32:21 WARN]:        at java.base/java.lang.Class.getConstructor(Class.java:2442)
[14:32:21 WARN]:        at packetevents-spigot-latest.jar//io.github.retrooper.packetevents.util.SpigotReflectionUtil.initConstructors(SpigotReflectionUtil.java:134)
[14:32:21 WARN]:        at packetevents-spigot-latest.jar//io.github.retrooper.packetevents.util.SpigotReflectionUtil.init(SpigotReflectionUtil.java:392)
[14:32:21 WARN]:        at packetevents-spigot-latest.jar//io.github.retrooper.packetevents.factory.spigot.SpigotPacketEventsBuilder$1.load(SpigotPacketEventsBuilder.java:103)
[14:32:21 WARN]:        at packetevents-spigot-latest.jar//io.github.retrooper.packetevents.factory.spigot.SpigotPacketEventsBuilder$1.init(SpigotPacketEventsBuilder.java:137)
[14:32:21 WARN]:        at packetevents-spigot-latest.jar//io.github.retrooper.packetevents.PacketEventsPlugin.onEnable(PacketEventsPlugin.java:47)
[14:32:21 WARN]:        at org.bukkit.plugin.java.JavaPlugin.setEnabled(JavaPlugin.java:287)
[14:32:21 WARN]:        at io.papermc.paper.plugin.manager.PaperPluginInstanceManager.enablePlugin(PaperPluginInstanceManager.java:202)
[14:32:21 WARN]:        at io.papermc.paper.plugin.manager.PaperPluginManagerImpl.enablePlugin(PaperPluginManagerImpl.java:109)
[14:32:21 WARN]:        at org.bukkit.plugin.SimplePluginManager.enablePlugin(SimplePluginManager.java:519)
[...]

Issue #764, opened 2024-05-10
NoSuchMethodException: net.minecraft.world.item.ItemStack.<init>(net.minecraft.world.item.Item,int)


RegotlyLong:
java.lang.NoSuchMethodException: net.minecraft.world.item.ItemStack.<init>(net.minecraft.world.item.Item,int)
[12:53:57 WARN]:        at java.base/java.lang.Class.getConstructor0(Class.java:3761)
[12:53:57 WARN]:        at java.base/java.lang.Class.getConstructor(Class.java:2442)
 at io.github.retrooper.packetevents.util.SpigotReflectionUtil.initConstructors(SpigotReflectionUtil.java:141)
[12:53:57 WARN]:        at io.github.retrooper.packetevents.util.SpigotReflectionUtil.init(SpigotReflectionUtil.java:360)
[12:53:57 WARN]:        at io.github.retrooper.packetevents.factory.spigot.SpigotPacketEventsBuilder$1.load(SpigotPacketEventsBuilder.java:103)
[12:53:57 WARN]:        at io.github.retrooper.packetevents.factory.spigot.SpigotPacketEventsBuilder$1.init(SpigotPacketEventsBuilder.java:137)
[12:53:57 WARN]:        at PacketEvents.getAPI().init();

java.lang.NullPointerException: Cannot invoke "java.lang.reflect.Field.get(Object)" because the return value of "com.github.retrooper.packetevents.util.reflection.Reflection.getField(java.lang.Class, java.lang.Class, int)" is null
        at io.github.retrooper.packetevents.util.SpigotReflectionUtil.getMinecraftServerConnectionInstance(SpigotReflectionUtil.java:406) 
        at io.github.retrooper.packetevents.injector.SpigotChannelInjector.isServerBound(SpigotChannelInjector.java:63) 
        at io.github.retrooper.packetevents.factory.spigot.SpigotPacketEventsBuilder$1.load(SpigotPacketEventsBuilder.java:115) 
        at io.github.retrooper.packetevents.factory.spigot.SpigotPacketEventsBuilder$1.init(SpigotPacketEventsBuilder.java:137) 

version 1.20.6 paper 
This server is running Paper version git-Paper-53 (MC: 1.20.6) (Implementing API version 1.20.6-R0.1-SNAPSHOT) (Git: 5729b29)

RegotlyLong:
[...]

PR #794, opened 2024-05-21
Add mojang mapped jar support

IssueDescription:
This adds support for mojang mapped jars, fixing #793 and #764


retrooper:
Thank you for this PR! Had to fix some conflicts, but should work out well.

PR #807, opened 2024-05-31
Removed: bStats Configuration Option

Bram1903:
## Explanation
We've observed that many users disable bStats for PacketEvents while keeping it enabled for their own plugins. This inconsistency raised questions about the necessity of having a separate bStats setting within PacketEvents. Since bStats can be disabled globally via `bStats/config.yml`, having an additional setting in PacketEvents was redundant and potentially confusing, as it allowed one plugin to enable bStats while another could disable it.

To streamline the configuration and reduce confusion, we have decided to remove the bStats configuration option accessible through `PacketEvents#getAPI#getSettings#bStats`. The relevant methods have been deprecated and rendered ineffective. Specifically, `PacketEventsSettings#bStats` will no longer have any effect, and `PacketEventsSettings#isbStatsEnabled` will always return true.

## Tasks
- [x] Mark the current methods as deprecated.
- [x] Made bStats specific methods ineffective. 
- [x] Remove internal bStats enabled checks to ensure bStats is always enabled.

## Clarification
For those concerned about this change, disabling bStats remains possible and is consistent with how you would disable it for other plugins. To disable bStats, navigate to the `bStats` directory within your server's plugins directory. In `config.yml`, you will find an option to disable bStats.


Aaron2404:
I love this change! The hypocrisy of the bstats situation, like you described, has always kind of bugged me.

> We've observed that many users disable bStats for PacketEvents while keeping it enabled for their own plugins.

retrooper (approved):
The description explains the situation. It is possible for your users to disable bStats, there's no necessity to disable it, specifically just for PacketEvents, on their behalf. And there's the hypocrisy, where some plugins might unintentionally disable it for PacketEvents, but enable it for their own plugin.

Issue #810, opened 2024-06-01
Treasure maps and other maps with icons cause a DecoderException when decoded as an item


joshuaprince:
**Describe the bug**
Reading any map that has an icon (most notably treasure maps) with wrappers such as WrapperPlayServerWindowItems kicks the player.

Client log:
```
[16:18:41] [Netty Epoll Client IO #13/ERROR]: Network Exception Caught
io.netty.handler.codec.DecoderException: Failed to decode packet 'clientbound/minecraft:container_set_content'
... truncated ...
Caused by: io.netty.handler.codec.DecoderException: Failed to decode: No key rotation in MapLike[{float:180.0f,type:"minecraft:red_x",x:-1399.0d,z:1545.0d}] missed input: {+:{float:180.0f,type:"minecraft:red_x",x:-1399.0d,z:1545.0d}} {+:{float:180.0f,type:"minecraft:red_x",x:-1399.0d,z:1545.0d}}
```

**Software brand**
Paper 1.20.6-130
PacketEvents 2.3.1-SNAPSHOT (rev 3352f25e)

**Plugins**
Just PacketEvents, built from source with added wrapper

**How To Reproduce**
```java
            public void onPacketPlaySend(PacketPlaySendEvent event) {
                if (event.getPacketType() == PacketType.Play.Server.WINDOW_ITEMS) {
                    WrapperPlayServerWindowItems w = new WrapperPlayServerWindowItems(event);
                    w.getItems();
                }
            }
```

Either find a shipwreck and open a chest that contains a treasure map, or give a map with an icon:
```
/give @s minecraft:map
[ explore the map ]
[ replace MAPID with the map ID just explored, and coordinates with your coordinates ]
/give @s minecraft:filled_map[minecraft:map_decorations={"a": {"type": "red_x", "x": MY_X, "z": MY_Z, "rotation": 0}}, minecraft:map_id=MAPID]
```

Log out and back in. Since the item is now in your inventory, it always kicks you on join until the WrapperPlayServerWindowItems is removed.

**Expected behavior**
No exception or kick.

**Screenshots**
N/A

**Additional context**
N/A

booky10:
Which client version? Also 1.20.5/1.20.6?

joshuaprince:
> Which client version? Also 1.20.5/1.20.6?

Also 1.20.6. Turned out to be a pretty simple fix, #811

PR #811, opened 2024-06-01
Fix map ItemStack decoding for rotation key

joshuaprince:
Fixes https://github.com/retrooper/packetevents/issues/810


booky10 (approved):
:+1:

AbhigyaKrishna:
Thank you for the contribution.

PR #812, opened 2024-06-02
Feat: Native Folia Support

Bram1903:
## Description
To further improve performance and prevent unnecessary usage of reflections, I have written the following PR. In this PR, I have switched from using the Spigot API to using the Paper API. Compatibility-wise, this change does not make any difference, but it now allows us to access the new threaded region schedulers natively.

## Tasks 
- [x] Switched from Spigot API to Paper API.
- [x] Updated Minecraft API version from 1.20.4 to 1.20.6.
- [x] Added `runVelocity` to both the version file and the Velocity implementation build file.
- [x] Introduced the `runFolia` task in the Spigot implementation.
- [x] Improved the `runServer` task in the Spigot implementation.
- [x] Refactored existing FoliaSchedulers to use the native API, eliminating reflections.
- [x] Updated `TaskWrapper` to utilize the `ScheduledTask` class directly instead of accessing it through reflections.
- [x] Updated GitHub build workflow to now use Java 21, instead of Java 17 (This is also required for Fabric and Sponge) 
- [x] Updated Travis to use Java 21, instead of Java 17 (This is also required for Fabric and Sponge) 

## Testing Results:

This PR has been tested on 1.8.8, 120.6, and Folia 1.20.6.

### Bukkit
image

### Folia
image


AbhigyaKrishna:
Please resolve the conflict from the latest commits

PR #819, opened 2024-06-05
Feat: Global Versioning

Bram1903:
# PacketEvents Global Versioning

This PR introduces a global versioning system for the entire library, aiming to simplify version management by removing the need to change versions across multiple files and locations.

## PEVersion
The `PEVersion` class has been completely rewritten to improve its implementation and support for the `SNAPSHOT` tag. The new class implements `Comparable<PEVersion>` and incorporates proper versioning logic. The previous integer array has been replaced with four fields: `int major`, `int minor`, `int patch`, and `boolean snapshot`.

### Key Changes:
- **Deprecated Constructor:** The old constructor is marked as deprecated.
- **New Constructors:** Two new constructors have been added. One requires semantic versioning, and the other is an overloaded method defaulting `snapshot` to false. These constructors throw an `IllegalArgumentException` if the provided version isn't in semantic versioning format.
- **SNAPSHOT Handling:** The `-SNAPSHOT` suffix is optional and defaults to false if not present.
- **Javadoc Updates:** The existing Javadoc has been revised to reflect the new implementation, with links to relevant classes and deprecated methods pointing to their replacements.
- **Comparable Implementation:** The class now implements `Comparable<PEVersion>`, enhancing readability and understanding of the version comparison logic.

### Versioning Example
- `2.3.1-SNAPSHOT` < `3.2.0`
- `2.3.1` > `2.3.1-SNAPSHOT`

`SNAPSHOT` is treated as a pre-release, thus `2.3.1` is considered newer than `2.3.1-SNAPSHOT`.

## Resource Processor
To streamline updating platform-specific version files (`plugin.yml`, `bungee.yml`, etc.), a task has been added to the `packetevents.library-convention`. This task replaces the version variable in these files during compilation, removing the need for manual updates.

### Velocity Implementation:
- The previous `@Plugin` annotation, which generated a `velocity-plugin.json` at runtime, has been removed.
- The `velocity-plugin.json` has now been manually placed in the resources directory with the version variable to be overwritten during the build process.

## Build Source PEVersion Task / Plugin
Initially, we considered adding the project version to the JAR manifest, but this approach proved unsuitable when shading the library into other plugins, as the manifest would be lost.

### Solution:
- A version file is created within the final build to indicate the library's version, ensuring compatibility even when the library is shaded.
- A new task, `PEVersionTask`, has been implemented by @AbhigyaKrishna. This task, named `generateVersionsFile`, creates a `PEVersions` file during compilation, instantiating the `PEVersion` class with the provided version.

This global versioning system simplifies version management and ensures consistent versioning across the entire library.

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/PacketEventsAPI.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/PacketEventsAPI.java
@@ -36,8 +36,8 @@ public abstract class PacketEventsAPI<T> {
     private static final PacketEventsSettings SETTINGS = new PacketEventsSettings();
     private static final UpdateChecker UPDATE_CHECKER = new UpdateChecker();
     private static final LogManager LOG_MANAGER = new LogManager();
+    private static final Logger LOGGER = Logger.getLogger(PacketEventsAPI.class.getName());
     private static final PEVersion VERSION = PEVersion.createFromPackageVersion();
-    private final Logger LOGGER = Logger.getLogger(PacketEventsAPI.class.getName());
 
     public EventManager getEventManager() {
         return EVENT_MANAGER;
--- a/api/src/main/java/com/github/retrooper/packetevents/PacketEventsAPI.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/PacketEventsAPI.java
@@ -27,6 +27,7 @@ import com.github.retrooper.packetevents.netty.NettyManager;
 import com.github.retrooper.packetevents.settings.PacketEventsSettings;
 import com.github.retrooper.packetevents.util.LogManager;
 import com.github.retrooper.packetevents.util.PEVersion;
+import com.github.retrooper.packetevents.util.PEVersions;
 import com.github.retrooper.packetevents.util.updatechecker.UpdateChecker;
 
 import java.util.logging.Logger;
@@ -37,7 +38,6 @@ public abstract class PacketEventsAPI<T> {
     private static final UpdateChecker UPDATE_CHECKER = new UpdateChecker();
     private static final LogManager LOG_MANAGER = new LogManager();
     private static final Logger LOGGER = Logger.getLogger(PacketEventsAPI.class.getName());
-    private static final PEVersion VERSION = PEVersion.createFromPackageVersion();
 
     public EventManager getEventManager() {
         return EVENT_MANAGER;
@@ -52,7 +52,7 @@ public abstract class PacketEventsAPI<T> {
     }
 
     public PEVersion getVersion() {
-        return VERSION;
+        return PEVersions.CURRENT;
     }
 
     public Logger getLogger() {
--- a/api/src/main/java/com/github/retrooper/packetevents/util/PEVersion.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/util/PEVersion.java
@@ -23,12 +23,16 @@ import org.jetbrains.annotations.NotNull;
 
 import java.util.Objects;
 import java.util.Optional;
+import java.util.logging.Logger;
 
 /**
  * Represents a PacketEvents version using Semantic Versioning.
  * Supports version comparison, cloning, and provides a string representation.
+ * Snapshots will always resolve to a newer version than the non-snapshot version if major, minor, and patch are the same.
  */
 public class PEVersion implements Comparable<PEVersion> {
+
+    public static final PEVersion UNKNOWN = new PEVersion(0, 0, 0);
 
     private final int major;
     private final int minor;
@@ -77,19 +81,31 @@ public class PEVersion implements Comparable<PEVersion> {
      *
      * @param version the version string (e.g., "1.8.9-SNAPSHOT").
      * @throws IllegalArgumentException if the version string format is incorrect.
+     * @deprecated use {@link #fromString(String)} instead.
      */
+    @Deprecated
     public PEVersion(@NotNull final String version) {
-        this.snapshot = version.endsWith("-SNAPSHOT");
         String versionWithoutSnapshot = version.replace("-SNAPSHOT", "");
         String[] parts = versionWithoutSnapshot.split("\\.");
 
-        if (parts.length != 3) {
-            throw new IllegalArgumentException("Version string must be in the format 'major.minor.patch[-SNAPSHOT]'");
+        if (parts.length < 2 || parts.length > 3) {
+            throw new IllegalArgumentException("Version string must be in the format 'major.minor[.patch][-SNAPSHOT]' found '" + version + "' instead.");
         }
 
         this.major = Integer.parseInt(parts[0]);
         this.minor = Integer.parseInt(parts[1]);
-        this.patch = Integer.parseInt(parts[2]);
+        this.patch = parts.length > 2 ? Integer.parseInt(parts[2]) : 0;
+        this.snapshot = version.contains("-SNAPSHOT");
+    }
+
+    /**
+     * Constructs a {@link PEVersion} instance from a version string.
+     *
+     * @param version the version string (e.g., "1.8.9-SNAPSHOT").
+     * @throws IllegalArgumentException if the version string format is incorrect.
+     */
+    public static PEVersion fromString(@NotNull final String version) {
+        return new PEVersion(version);
     }
 
     /**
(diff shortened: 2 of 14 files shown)


buildSrc/src/main/kotlin/packetevents.library-conventions.gradle.kts
```diff
@@ -30,6 +30,12 @@ tasks {
         options.release = 8
     }
 
+    processResources {
+        filesMatching(listOf("plugin.yml", "bungee.yml", "velocity-plugin.json", "fabric.mod.json")) {
```
booky10, line 34: The inputs should be set here to make gradle not use the cache when just changing the version

api/src/main/java/com/github/retrooper/packetevents/PacketEventsAPI.java
```diff
@@ -35,11 +35,16 @@ public abstract class PacketEventsAPI<T> {
...
-    private static final PEVersion VERSION = new PEVersion(2, 3, 1);
-    
+    private final Logger LOGGER = Logger.getLogger(PacketEventsAPI.class.getName());
+
+    private static final String version = PacketEvents.class.getPackage().getImplementationVersion();
+
+    // The version needs to be checked for null, because when running Unit Tests, the version will be null.
+    private static final String[] versionParts = (version != null) ? version.split("-") : new String[]{ "0.0.0" };
+    private static final String[] parts = versionParts[0].split("\\.");
+    private static final PEVersion VERSION = new PEVersion(Integer.parseInt(parts[0]), Integer.parseInt(parts[1]), Integer.parseInt(parts[2]));
```
booky10, line 46: Imo this should be implemented as a static method in the PEVersion class

api/src/main/java/com/github/retrooper/packetevents/PacketEventsAPI.java
```diff
@@ -35,11 +35,16 @@ public abstract class PacketEventsAPI<T> {
     private static final EventManager EVENT_MANAGER = new EventManager();
     private static final PacketEventsSettings SETTINGS = new PacketEventsSettings();
     private static final UpdateChecker UPDATE_CHECKER = new UpdateChecker();
-    private final Logger LOGGER = Logger.getLogger(PacketEventsAPI.class.getName());
     private static final LogManager LOG_MANAGER = new LogManager();
-    //TODO UPDATE
-    private static final PEVersion VERSION = new PEVersion(2, 3, 1);
-    
+    private final Logger LOGGER = Logger.getLogger(PacketEventsAPI.class.getName());
```
booky10, line 39: Is there a reason to this being moved down one line?

api/src/main/java/com/github/retrooper/packetevents/PacketEventsAPI.java
Bram1903: Will do, thanks for the feedback!

buildSrc/src/main/kotlin/packetevents.library-conventions.gradle.kts
Bram1903: I am not entirely sure what you mean here? Could you elaborate?

api/src/main/java/com/github/retrooper/packetevents/PacketEventsAPI.java
Bram1903: This has now been implemented. Can you take a look at the latest commit?

AbhigyaKrishna:
You could also rework the `PEVersion` class. As we are always following the `major.minor.patch-SNAPSHOT` semantic, the class could contain there integers and a boolean for snapshot. And also can implement `Comparable<PEVersion>` to make it consistent.

buildSrc/src/main/kotlin/packetevents.library-conventions.gradle.kts
booky10: just so `inputs.property("version", project.version)` before this line

api/src/main/java/com/github/retrooper/packetevents/PacketEventsAPI.java
booky10: oh nvm, i assumed it was static because of the naming

api/src/main/java/com/github/retrooper/packetevents/PacketEventsAPI.java
booky10: looks good 👍

Bram1903:
> You could also rework the `PEVersion` class. As we are always following the `major.minor.patch-SNAPSHOT` semantic, the class could contain there integers and a boolean for snapshot. And also can implement `Comparable<PEVersion>` to make it consistent.

Sure, thanks for the tip! Will do!

buildSrc/src/main/kotlin/packetevents.library-conventions.gradle.kts
Bram1903: Should be good now, right?

Bram1903:
Hey @AbhigyaKrishna,

I just finished rewriting the `PEVersion` class. Could you take a look at it? It now implements both `Comparable` and `Cloneable`.
Personally, I think this is a much-improved version of the previous implementation. 

It now also has proper snapshot support, and I think it's overall more useable.

AbhigyaKrishna (changes requested):

AbhigyaKrishna (approved):
Looks good

api/src/main/java/com/github/retrooper/packetevents/util/PEVersion.java
```diff
@@ -18,142 +18,201 @@
...
     /**
-     * Specify your version using an array.
+     * Constructs a {@link PEVersion} instance with snapshot defaulted to false.
      *
-     * @param version Array version.
+     * @param major the major version number.
+     * @param minor the minor version number.
+     * @param patch the patch version number.
      */
-    public PEVersion(final int... version) {
```
booky10, line 42: this constructor should be re-added, otherwise this would be a breaking change

Bram1903: > this constructor should be re-added, otherwise this would be a breaking change

Shall I mark it as deprecated, and then make it use the overload constructor which defaults snapshot to false?

booky10: Oops, missed your response - yeah, imo thats a good solution

api/src/main/java/com/github/retrooper/packetevents/util/PEVersion.java
```diff
@@ -18,142 +18,212 @@
...
     }
 
     /**
-     * Represented as an array.
+     * Checks if the version is a snapshot.
      *
-     * @return Array version.
+     * @return true if snapshot, false otherwise.
      */
-    public int[] asArray() {
```
booky10, line 114: this method has been removed, is this an intended breaking change?

Bram1903: Yes, I wasn't sure how to properly implement it without just skipping the snapshot part. Do you want me to add it, and mark it as deprecated? This would mean snapshot won't be accounted for in the return value of that method.

booky10: :+1: snapshot wasn't present there previously either, an `@Deprecated` is definetely needed

Bram1903: I'll add this tomorrow morning, and test this while shading PE as well!

booky10: You've called it "toArray", it should be named "asArray" 😄

AbhigyaKrishna:
We are now storing the current version as a constant. This file is generated by a gradle task is generated every build.

api/src/main/java/com/github/retrooper/packetevents/util/PEVersion.java
Bram1903: Fixed 😅

Bram1903:
LGTM

booky10 (approved):
:+1:

Issue #821, opened 2024-06-06
WrapperPlayServerChunkDataBulk not initialized correctly when instantiated multiple times


RaynLegends:
**Describe the bug**
If a plugin calls `new WrapperPlayServerChunkDataBulk(event)` the wrapper is initialized correctly the first time it gets called (each field has the correct non-null values).
Subsequent calls create WrapperPlayServerChunkDataBulk objects with every field set to null.

**Software brand**
Tested in Paper 1.8.8 and Pufferfish 1.20.4

**Plugins**
A single plugin with the latest version of PacketEvents shaded

**How To Reproduce**
1. Create a plugin that listens multiple times to PacketPlaySendEvent MULTI_BLOCK_CHANGE
2. Call new WrapperPlayServerChunkDataBulk(event)
3. Try to access a field, for example calling WrapperPlayServerChunkDataBulk#getX

**Expected behavior**
WrapperPlayServerChunkDataBulk#getX and every other method should return non-null values

**Screenshots**
WrapperPlayServerChunkDataBulk#getX returns an array only for the first object created, returns null otherwise

**Additional context**
Looking at the code, it seems that WrapperPlayServerChunkDataBulk does not override PacketWrapper#copy

PR #822, opened 2024-06-06
Fix: WrapperPlayServerChunkDataBulk not initialized correctly when instantiated multiple times

RaynLegends:
Fixes retrooper/packetevents#821


retrooper:
Thanks for the PR.

PR #835, opened 2024-06-16
Divide Color's int rgb by 255 when using for particle colors

libraryaddict:
Currently it'll use the int as is from Color, eg: `0 to 255`.
It should be a decimal from `0.0 to 1.0` as per https://wiki.vg/Particles


booky10 (approved):

booky10:
thanks for the pr!

PR #846, opened 2024-06-23
Have fromBukkitEntityType invoke getByLegacyId instead

libraryaddict:
Currently it's invoking getById() which is 1.14+, should be getByLegacyId()


libraryaddict:
So apparently switching it will break other entity types, like Minecart. I cba thinking about this

PR #847, opened 2024-06-23
Fix: Netty Version not being properly converted to an PEVersion

Bram1903:
Make sure to handle NettyVersions properly as they don't follow semantic versioning.
This fixes the following issue:
image

PS. Bumped the runServer task to use Minecraft server version 1.21 while I was at it.


retrooper:
Thanks.

Issue #645, opened 2024-01-04
WrapperPlayServerChatMessage returns incorrect ChatType for certain commands


KaspianDev:
**Describe the bug**
WrapperPlayServerChatMessage returns incorrect ChatType for some commands (like /say). Mappings could be incorrect or outdated (latest chat type mappings are for 1.19.1, protocol might've changed since).

**Software brand**
paper 1.20.2

**Plugins**
worldedit, worldguard, vault, luckperms, placeholderapi, antipopup (this one uses packetevents), skinsrestorer

**How To Reproduce**
1. Create a PacketListener like so:
```java
public class SomeListener extends PacketListenerAbstract {

    @Override
    public void onPacketSend(PacketSendEvent event) {
        if (event.getPacketType() == PacketType.Play.Server.CHAT_MESSAGE) {
            WrapperPlayServerChatMessage wrapper = new WrapperPlayServerChatMessage(event);

            ChatType chatType = wrapper.getMessage().getType();
            System.out.println(chatType.getId(event.getUser().getClientVersion()));
            System.out.println(chatType.getName());
        }
    }
 
}
```

2. Join the game and execute for example /say command.
3. You will notice that the console outputs this:
```
[16:17:17 INFO]: USERNAME issued server command: /say hello
[16:17:17 INFO]: [Not Secure] [USERNAME] hello
[16:17:17 INFO]: 5
[16:17:17 INFO]: minecraft:team_msg_command_outgoing
```

**Expected behavior**
Correct ChatType should be returned (minecraft:say_command).
Latest mappings say that ID 1 translates to say_command so either wrong ID is returned or mappings are wrong/outdated.

PR #852, opened 2024-06-27
feat: chat type mappings update

AbhigyaKrishna:
fixes: #645


retrooper:
What about "raw" in 1.19+ mappings?

AbhigyaKrishna:
It was removed!
Source: https://gist.github.com/nikes/aff59b758a807858da131a1881525b14

retrooper:
Thank you!

PR #860, opened 2024-06-28
PR change the location of bytebuf modification methods from PacketWrapper to a seperate exposed API

Tofaa2:
This PR aims to replace the PacketWrapper#readSomething and PacketWrapper#writeSomething with a more scalable and convenient API. Seperating types and no longer depending on Object for everything in PacketWrapper that is meant to be ByteBuf (kinda)


retrooper (approved):

sh2kcs:
When will this be added?

retrooper:
It's not complete, and it's still marked as a draft?

Tofaa2:
Pr back to being actively worked on

api/src/main/java/com/github/retrooper/packetevents/binary/BinaryBuffer.java
```diff
@@ -227,6 +236,53 @@ public <T> T read(BinaryBufferType<T> type) {
         return type.read(this);
     }
 
+
```
Tofaa2, line 240: @retrooper Do you think these should be kept as methods in here or static BinaryBufferType<?> creation methods. Same with map and collection and primitive arrays with varint size prefix

PR #865, opened 2024-06-30
Switch EventManager listenerSet to CopyOnWriteArraySet

cswhite2000:
The current ConcurrentHashMap.newKeySet in EventManager can cause fairly significant overhead in thread usage.

The [CopyOnWriteArraySet](https://docs.oracle.com/javase%2F8%2Fdocs%2Fapi%2F%2F/java/util/concurrent/CopyOnWriteArraySet.html) is generally only preferred when you have a very small number of writes and a very high number of reads. The usage here is one of these cases, as it should be rather uncommon for plugins to remove packet listeners. On the other hand, this set is iterated over every single time a packet is sent, which is incredibly common.

I've tested this using Grim on a server that averages 60-80 concurrent players, and observed significantly lower netty thread usage.
I don't currently have the spark flame graph handy as I've had this change running for several months. I can replicate a spark report with this data if required.


retrooper:
Thank you for the change!

Issue #857, opened 2024-06-28
1.12.2 Spigot memory leak `java.lang.OutOfMemoryError: Java heap space`


DreamVoid:
**Describe the bug**
```
...server loading without errors...
[15:36:03 INFO]: [packetevents] Loading packetevents v2.3.0
[15:36:08 ERROR]: [org.bukkit.craftbukkit.v1_12_R1.CraftServer] Java heap space initializing packetevents v2.3.0 (Is it up to date?)
java.lang.OutOfMemoryError: Java heap space
        at java.io.DataInputStream.readUTF(Unknown Source) ~[?:?]
        at java.io.DataInputStream.readUTF(Unknown Source) ~[?:?]
        at com.github.retrooper.packetevents.protocol.nbt.serializer.DefaultNBTSerializer$$Lambda$429/0x00007f983073a590.readName(Unknown Source) ~[?:?]
        at com.github.retrooper.packetevents.protocol.nbt.serializer.NBTSerializer.readTagName(NBTSerializer.java:107) ~[?:?]
        at com.github.retrooper.packetevents.protocol.nbt.serializer.DefaultNBTSerializer.lambda$new$22(DefaultNBTSerializer.java:97) ~[?:?]
        at com.github.retrooper.packetevents.protocol.nbt.serializer.DefaultNBTSerializer$$Lambda$453/0x00007f9830778678.readTag(Unknown Source) ~[?:?]
        at com.github.retrooper.packetevents.protocol.nbt.serializer.NBTSerializer.readTag(NBTSerializer.java:115) ~[?:?]
        at com.github.retrooper.packetevents.protocol.nbt.serializer.DefaultNBTSerializer.lambda$new$24(DefaultNBTSerializer.java:121) ~[?:?]
        at com.github.retrooper.packetevents.protocol.nbt.serializer.DefaultNBTSerializer$$Lambda$455/0x00007f9830778ad8.readTag(Unknown Source) ~[?:?]
        at com.github.retrooper.packetevents.protocol.nbt.serializer.NBTSerializer.readTag(NBTSerializer.java:115) ~[?:?]
        at com.github.retrooper.packetevents.protocol.nbt.serializer.DefaultNBTSerializer.lambda$new$22(DefaultNBTSerializer.java:97) ~[?:?]
        at com.github.retrooper.packetevents.protocol.nbt.serializer.DefaultNBTSerializer$$Lambda$453/0x00007f9830778678.readTag(Unknown Source) ~[?:?]
[...]

PR #870, opened 2024-07-02
Mappings reading optimization

AbhigyaKrishna:
This PR introduces a sequential pattern for deserializing mappings. Instead of converting the whole stream to an NBT object at once then deserializing the nbt object, we have an approach of directly converting data from the streams to our mappings objects without any intermediate NBT object.

# Previous Approach
image

# Current Approach
image

# Changelogs
* [x] Rework NBTLimiter class with better code design
* [x] NBTSerializer read functions now have package visibility
* [x] SequentialNBTReader implementation (reads data stream one at a time)
* [x] Mappings format v2

# Issues
#857


Bram1903:
I am always amazed how you, and Retrooper and Booky for that matter, come up with ideas and improvements like these. This looks like a great improvement!

Bram1903:
> Nice work. Did someone run a profiler and compared both versions yet?

Not sure, but feel free to do so :-)

SquidXTV:
So I did a quick check and don't see that much of an improvement. 
I tested it using a fresh `1.21.0` spigot server and build both PE jars and used the `2.4.1-SNAPSHOT` version.
Additionally, I used intellijs profiler and with this run configuration:

run config

I focused on the `PacketEventsPlugin#onLoad` method and its memory/cpu usage:

## Before
CPU time = `1674ms`
Memory allocations = `691.33 MB`

## After
CPU time = `1807ms`
Memory allocations = `668.01 MB`

Here are the full JFR snapshots generated:
[snapshots.zip](https://github.com/user-attachments/files/16109649/snapshots.zip)

SquidXTV:
I just checked again with VisualVM and loading the .jfr snapshots and could find out that the heap size dropped.
So it seems like the peak memory usage is a smaller:

## Before
before

## After
after

AbhigyaKrishna:
That's some great news!

SquidXTV:
> That's some great news! Most of the objects are dropped after a single read but Java GC doesn't clear it at the very moment. It would be interested to see how much overall memory gain this has!

Overall they kinda use the same amount of memory, but the new versions seem to get rid of not needed stuff earlier, making the peak memory usage of that part smaller.

## Before
before

## After
after

SquidXTV:
Doesnt look like that much of a change tbh.

## Second Version
memory

intellij

AbhigyaKrishna:
We actually improved the NBTCompound caching part!
image

There are no more hashmaps getting created in the memory while loading block mappings.

retrooper:
Thank you for this clever optimization @AbhigyaKrishna!

PR #872, opened 2024-07-02
Optimized the EventManager

ShadowOfHeaven-Me:
The full description of what was changed and the reasons why can be found inside the class. If you have any additional questions or reservations, feel free to ask.


api/src/main/java/com/github/retrooper/packetevents/event/EventManager.java
```diff
@@ -96,28 +116,55 @@ public PacketListenerCommon registerListener(PacketListenerCommon listener) {
...
     }
 
 
     /**
      * Unregister all dynamic packet event listeners.
      */
     public void unregisterAllListeners() {
-        listenersMap.clear();
+        this.listenersMap.clear();
+        this.listeners = new PacketListenerCommon[0];
```
booky10, line 143: Shouldnt this call the recalc method to prevent race conditions?

booky10: You could also just completely replace this implementation:
```java
this.unregisterListeners(this.listeners);
```

booky10 (changes requested):
Looks good aside from one small edge-case 👍

api/src/main/java/com/github/retrooper/packetevents/event/EventManager.java
ShadowOfHeaven-Me: > Shouldnt this call the recalc method to prevent race conditions?

Oh, right. Both of these are valid solutions. Another possible solution is to add a synchronized(this) block on the update of 'listeners' (both in recalculate and unregisterAll). I think that one would be the fastest, so I'd go with that

retrooper:
Any numbers? Comparing the previous code with this?

ShadowOfHeaven-Me:
> Any numbers? Comparing the previous code with this?

You can test for yourself if you're curious how much faster the iteration is

ShadowOfHeaven-Me:
> Any numbers? Comparing the previous code with this?

I just made a primitive but good enough benchmark.
As per this code: https://paste.md-5.net/ofusemesow.java
The results of the optimized iteration were about ~25 times faster than the previous code (old code scored on average 0.55 ms, while the new code about 0.002 ms).
The test was done with 6 listeners on 3 different priorities

retrooper:
The user storage system has nothing to do with this PR. It should be a separate PR / issue.

ShadowOfHeaven-Me:
> The user storage system has nothing to do with this PR. It should be a separate PR / issue.

Yeah, that's what I planned to do, but it automatically merged itself. I'm not sure how to reverse it now

booky10 (approved):
Although there still is the possibility of the listeners map changing while rebuilding the listeners array causing new and outdated information to mix, this doesn't seem like a huge problem to me.
Overall, looks good :+1:

ShadowOfHeaven-Me:
"map changing while rebuilding" Wouldn't another listeners recalculation call be waiting to update the listeners to the most up-to-date value? I don't see how the Map could be changed with without the listeners containg exactly what the map contains

retrooper:
Why has this been closed?

ShadowOfHeaven-Me:
Well I thought that if it was approved, there was no need for the PR to stay open

Bram1903:
> Well I thought that if it was approved, there was no need for the PR to stay open

There is a difference between a single collaborator approving a PR, and it actually being merged ;-)

retrooper:
Thank you for the PR!

ShadowOfHeaven-Me:
Yeah np. Thanks for merging. I'm gonna upload the storage system tomorrow

PR #877, opened 2024-07-08
Add current commit hash to project version

booky10:
Only added when version is a snapshot version, disables itself if no git directory is found

---

Shortly discussed [on discord](https://canary.discord.com/channels/721686193061888071/782010220305973270/1257279805486006273), this allows to see which packetevents build a user has installed

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/util/PEVersion.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/util/PEVersion.java
@@ -180,8 +180,8 @@ public class PEVersion implements Comparable<PEVersion> {
 
     /**
      * Gets the snapshot commit hash. May be of any length.
-     * Availability is not guaranteed.
-     * Additionally, its presence is not contingent on this being a snapshot version.
+     * Availability is not guaranteed since it is contingent on how the program was built.
+     * Generally speaking, the commit hash is only available if the version is a snapshot version.
      *
      * @return the snapshot commit hash, if available.
      */
--- a/api/src/main/java/com/github/retrooper/packetevents/util/PEVersion.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/util/PEVersion.java
@@ -283,7 +283,8 @@ public class PEVersion implements Comparable<PEVersion> {
 
     /**
      * Converts the {@link PEVersion} to an array of version numbers.
-     * @deprecated since {@link PEVersion} instances now always use Semantic Versioning.
+     * @deprecated since {@link PEVersion} instances now always use Semantic Versioning, and thus, returning an "amorphous" array,
+     * with no definitive size, seems redundant.
      * @return an array of version numbers.
      */
     @Deprecated
--- a/api/src/main/java/com/github/retrooper/packetevents/util/PEVersion.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/util/PEVersion.java
@@ -179,9 +179,9 @@ public class PEVersion implements Comparable<PEVersion> {
     }
 
     /**
-     * Gets the snapshot commit hash. May be of any length.
+     * Gets the snapshot commit hash of the PacketEvents snapshot version. May be of any length.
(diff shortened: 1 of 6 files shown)


Bram1903:
Shouldn't we add a getter for the commit hash in the `PEVersion` class? This would then simply return an empty string if the snapshot is set to false, but otherwise return the commit hash.

Bram1903:
> As far as I know the SNAPSHOT has to be at the end because of maven Appending the hash with a + is also valid semver, thats why I did that

Got it! Thanks for adding the getter!

Bram1903 (approved):
LGTM

api/src/main/java/com/github/retrooper/packetevents/util/PEVersion.java
```diff
@@ -26,7 +26,9 @@
 /**
  * Represents a PacketEvents version using Semantic Versioning.
  * Supports version comparison, cloning, and provides a string representation.
- * Snapshots will always resolve to a newer version than the non-snapshot version if major, minor, and patch are the same.
```
retrooper, line 28: This was not accurate and was corrected.

api/src/main/java/com/github/retrooper/packetevents/util/PEVersion.java
```diff
@@ -178,8 +180,8 @@ public boolean snapshot() {
 
     /**
      * Gets the snapshot commit hash. May be of any length.
-     * May always be present or absent, doesn't matter
-     * wether this is a snapshot version or not.
```
retrooper: Typo: "wether".
Claiming that the value is present or absent sounds pretty redundant, so this was already rephrased.

api/src/main/java/com/github/retrooper/packetevents/util/PEVersion.java
```diff
@@ -191,7 +193,7 @@ public boolean snapshot() {
      * Compares this {@link PEVersion} with another {@link PEVersion}.
      *
      * @param other the other {@link PEVersion}.
-     * @return a negative integer, zero, or a positive integer as this version is less than,
```
retrooper, line 146: A little confusing on the first read.

api/src/main/java/com/github/retrooper/packetevents/util/PEVersion.java
```diff
@@ -281,7 +283,7 @@ public String toString() {
 
     /**
      * Converts the {@link PEVersion} to an array of version numbers.
-     *
+     * @deprecated since {@link PEVersion} instances now always use Semantic Versioning.
```
retrooper, line 286: Explaining why this method was deprecated.

api/src/main/java/com/github/retrooper/packetevents/util/PEVersion.java
```diff
@@ -283,7 +283,8 @@ public String toString() {
 
     /**
      * Converts the {@link PEVersion} to an array of version numbers.
-     * @deprecated since {@link PEVersion} instances now always use Semantic Versioning.
+     * @deprecated since {@link PEVersion} instances now always use Semantic Versioning, and thus, returning an "amorphous" array,
```
retrooper, line 286: Proper explanation for why the PEVersion#asArray method was deprecated.

api/src/main/java/com/github/retrooper/packetevents/util/PEVersion.java
```diff
@@ -139,11 +178,22 @@ public boolean snapshot() {
         return snapshot;
     }
 
+    /**
+     * Gets the snapshot commit hash. May be of any length.
+     * Availability is not guaranteed since it is contingent on how the program was built.
+     * Generally speaking, the commit hash is only available if the version is a snapshot version.
```
retrooper, line 184: More clarity in documentation now.

retrooper (approved):
LGTM 👍

retrooper:
Thank you.

PR #881, opened 2024-07-10
Don't specify version hash in published maven pom

booky10:
Fixes dependency resolution errors since #877


Bram1903:
Let's go! 💪🏻

PR #882, opened 2024-07-10
Fix: Palette index calculation error

### LINKED ISSUE #880: Biome palette index is not calculated correctly
**Describe the bug**
The maximum size of the biome palette is (4\*4\*4)-1, but the packet events only calculates the index for a (16\*16\*16)-1 size palette, causing this bug.

**Software brand**
Paper 1.20.6

**Plugins**
PacketEvents, TestPlugin

**How To Reproduce**
1. Try this code or [plugin](https://github.com/user-attachments/files/16162521/plugin.zip)

```java
DataPalette palette = DataPalette.createForBiome();
for (int x = 0; x < 4; x++) {     
    for (int y = 0; y < 4; y++) {
        for (int z = 0; z < 4; z++) {
            palette.set(x, y, z, 0);
        }
    }
}
```
2. Start the server
3. See error

**Expected behavior**
Calculate the index correctly

**Screenshots**

screenshot

**Additional context**
N/A

KoutaChan:
fixes #880

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/protocol/world/chunk/palette/DataPalette.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/protocol/world/chunk/palette/DataPalette.java
@@ -187,6 +187,6 @@ public class DataPalette {
     }
 
     private static int index(PaletteType paletteType, int x, int y, int z) {
-        return (y << paletteType.getBitShift() | z) | x << paletteType.getBitShift();
+        return (y << paletteType.getBitShift() | z) << x | paletteType.getBitShift();
     }
 }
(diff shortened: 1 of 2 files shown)


api/src/main/java/com/github/retrooper/packetevents/protocol/world/chunk/palette/DataPalette.java
```diff
@@ -186,7 +186,7 @@ private static Palette createPalette(int bitsPerEntry, PaletteType paletteType)
         }
     }
 
-    private static int index(int x, int y, int z) {
-        return y << 8 | z << 4 | x;
+    private static int index(PaletteType paletteType, int x, int y, int z) {
+        return (y << paletteType.getBitShift() | z) | x << paletteType.getBitShift();
```
booky10, line 190: this is incorrect, y should be shifted two times the bit shift
otherwise y and x would "occupy" the same bits of the index

booky10 (changes requested):

api/src/main/java/com/github/retrooper/packetevents/protocol/world/chunk/palette/DataPalette.java
KoutaChan: Yes, I made a mistake. I pushed for a commit and would you please check it.

booky10 (approved):
:+1:

PR #885, opened 2024-07-10
Implement automatic javadoc publishing

booky10:
~~Works on my fork (see https://booky10.github.io/packetevents/) but I'm not sure how this should get deployed, so this is a draft PR for now~~

See also https://github.com/packetevents/packetevents.github.io/pull/1


retrooper:
https://github.com/packetevents

PR #886, opened 2024-07-11
User Storage System

ShadowOfHeaven-Me:
Added a system that allows developers to associate a certain value with the PacketEvent User class, making it easier and faster to identify the user on any PacketEvent at all times. MethodHandle was used instead of VarHandle, as I found it to be slightly faster, with, and without warmup calls. Here's the benchmark with 1000 warmup calls and 10000 evaluation calls: https://paste.md-5.net/exedugixog.cpp

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/event/EventManager.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/event/EventManager.java
@@ -20,6 +20,7 @@ package com.github.retrooper.packetevents.event;
 
 import com.github.retrooper.packetevents.PacketEvents;
 import com.github.retrooper.packetevents.exception.InvalidHandshakeException;
+import com.github.retrooper.packetevents.protocol.packettype.PacketType;
 import org.jetbrains.annotations.Nullable;
 
 import java.util.*;
--- a/api/src/main/java/com/github/retrooper/packetevents/event/EventManager.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/event/EventManager.java
@@ -20,18 +20,20 @@ package com.github.retrooper.packetevents.event;
 
 import com.github.retrooper.packetevents.PacketEvents;
 import com.github.retrooper.packetevents.exception.InvalidHandshakeException;
-import com.github.retrooper.packetevents.protocol.packettype.PacketType;
 import org.jetbrains.annotations.Nullable;
 
-import java.util.*;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.logging.Level;
 
 /**
- * Class for event managing. Implements both, internal and API methods.
+ * Class for event management. Implements both, internal and API methods.
  *
- * @apiNote It is highly recommended to register the event listeners at server start-up,
+ * @apiNote It is highly recommended to register the event listeners during server start-up,
  * without frequent modifications during runtime, since this class needs to recalculate all of
  * its event listeners during any modifications. In other words, it assumes that reads (events fired)
  * greatly outnumber writes (listeners modifications). If your case requires frequent listener modifications,
--- a/api/src/main/java/com/github/retrooper/packetevents/protocol/player/User.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/protocol/player/User.java
@@ -390,7 +390,7 @@ public class User {
      *
      */
 
-    public void store(TypedStorageValueId<?> identifier, Object value) {
+    public <T> void store(TypedStorageValueId<T> identifier, T value) {
         this.fastUserStorage.store(identifier, value);
     }
 
--- a/api/src/main/java/com/github/retrooper/packetevents/protocol/player/User.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/protocol/player/User.java
@@ -19,7 +19,6 @@
 package com.github.retrooper.packetevents.protocol.player;
 
 import com.github.retrooper.packetevents.PacketEvents;
-import com.github.retrooper.packetevents.event.EventManager;
 import com.github.retrooper.packetevents.manager.server.ServerVersion;
 import com.github.retrooper.packetevents.netty.channel.ChannelHelper;
 import com.github.retrooper.packetevents.protocol.ConnectionState;
@@ -47,11 +46,8 @@ import net.kyori.adventure.text.Component;
 import org.jetbrains.annotations.Nullable;
 
 import java.net.InetSocketAddress;
-import java.util.HashMap;
 import java.util.List;
-import java.util.TreeMap;
 import java.util.UUID;
-import java.util.concurrent.ConcurrentHashMap;
 
 public class User {
     private final Object channel;
@@ -380,7 +376,7 @@ public class User {
      * @param value The value to store.
      */
 
-    public void storeRuntime(StorageValueId identifier, Object value) {
+    public void store(StorageValueId identifier, Object value) {
         this.fastUserStorage.store(identifier, value);
     }
 
@@ -394,7 +390,7 @@ public class User {
      *
      */
 
-    public void storeRuntime(TypedStorageValueId<?> identifier, Object value) {
+    public void store(TypedStorageValueId<?> identifier, Object value) {
         this.fastUserStorage.store(identifier, value);
     }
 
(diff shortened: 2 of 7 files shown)


api/src/main/java/com/github/retrooper/packetevents/event/EventManager.java
```diff
@@ -20,6 +20,7 @@
 
 import com.github.retrooper.packetevents.PacketEvents;
 import com.github.retrooper.packetevents.exception.InvalidHandshakeException;
+import com.github.retrooper.packetevents.protocol.packettype.PacketType;
```
booky10, line 23: any reason for this random import being added here?

api/src/main/java/com/github/retrooper/packetevents/protocol/player/User.java
```diff
@@ -361,4 +369,52 @@ public NBTCompound getWorldNBT(String worldName) {
...
+     * Stores the given {@param value} with the given {@param identifier}.
+     * This essentially makes this User object associated with the stored value.
+     * Enables a quick look-up of the associated value with the {@link #getStored(StorageValueId)}
+     * method.
+     *
+     * @param identifier The plugin's constant identifier created from {@link StorageValueId#identifierFor StorageValueId.identifierFor}
+     * @param value The value to store.
+     */
+
+    public void storeRuntime(StorageValueId identifier, Object value) {
```
booky10, line 383: i think just `store` would be sufficient for this method name

api/src/main/java/com/github/retrooper/packetevents/protocol/player/User.java
```diff
@@ -361,4 +369,52 @@ public NBTCompound getWorldNBT(String worldName) {
...
+     * Stores the given {@param value} with the given {@param identifier}.
+     * This essentially makes this User object associated with the stored value.
+     * Enables a quick look-up of the associated value with the {@link #getStored(TypedStorageValueId)}
+     * method.
+     *
+     * @param identifier The plugin's constant identifier created from {@link TypedStorageValueId#identifierFor TypedStorageValueId.identifierFor}
+     *
+     */
+
+    public void storeRuntime(TypedStorageValueId<?> identifier, Object value) {
```
booky10, line 397: see above, also any reason for not using generics here? would ensure more type-safety at compile-time

api/src/main/java/com/github/retrooper/packetevents/protocol/player/storage/FastUserStorage.java
```diff
@@ -0,0 +1,112 @@
...
+        return this.get0(identifier.id());
+    }
+
+    public <T> T get(TypedStorageValueId<T> identifier) {
+        return (T) this.get0(identifier.id());
+    }
+
+    private Object get0(int index) {
+        if (index >= data.length) {
+            this.regrow();
```
booky10, line 74: why are you growing here?

api/src/main/java/com/github/retrooper/packetevents/protocol/player/storage/FastUserStorage.java
```diff
@@ -0,0 +1,112 @@
...
+ * @author ShadowOfHeaven
+ */
+
+public final class FastUserStorage {
+
+    //Ensure the array's element thread visibility by using MethodHandles
+    //It was found to be slightly faster than VarHandles with excessive amount of calls
+    private static final MethodHandle setElement = MethodHandles.arrayElementSetter(Object[].class);//array, index, value
+    private static final MethodHandle getElement = MethodHandles.arrayElementGetter(Object[].class);//array, index
+    private static final Map<Object, Integer> pluginToId = new ConcurrentHashMap<>(2);
```
booky10, line 43: if the keys here are plugin instances, this should definetely be a `WeakHashMap`
otherwhise this will cause memory leaks when a plugin gets disabled or unloaded

because of this you will need to do manual synchronization, see some coment below for mor info

api/src/main/java/com/github/retrooper/packetevents/protocol/player/storage/FastUserStorage.java
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    private void store0(int index, Object value) {
+        if (index >= data.length) this.regrow();
+        try {
+            setElement.invokeExact(this.data, index, value);//must be invoked with an int
+        } catch (Throwable e) {
+            throw new RuntimeException(e);
+        }
+    }
```
booky10, line 62: this is definetely not thread safe. imagine following scenario: two threads want to store something concurrently.
both start at the same time, the first one grows the data array and then sets the element in the data array
the problem now is that if the second thread also starts growing the array, the data of the first thread will be completely discarded

---

this can be solved by locking the array while growing and also while setting an element
im sure this is possible using some methodhandle magic, but i would suggest using an `ReadWriteLock` (implementation `ReentrantReadWriteLock`) if you don't know any better

using this lock, you can lock the write lock while storing something and lock the read lock while getting something
the `ReadWriteLock` manages these locks so the read lock can be locked multiple times, while the write lock blocks everything

api/src/main/java/com/github/retrooper/packetevents/protocol/player/storage/FastUserStorage.java
booky10: im also not sure if its a good idea to key these ids by "plugin" instance - in my opinion it makes more sense to key the id map by the class of the identifier type, so you can create multiple keys per plugin

api/src/main/java/com/github/retrooper/packetevents/protocol/player/storage/StorageValueId.java
```diff
@@ -0,0 +1,26 @@
+package com.github.retrooper.packetevents.protocol.player.storage;
+
+public final class StorageValueId extends StorageIdBase {
```
booky10, line 3: is there any reason for this to exist?

booky10 (changes requested):
aside from the comments, your javadoc formatting is a bit inconsistent - but that's just a minor nitpick

api/src/main/java/com/github/retrooper/packetevents/event/EventManager.java
ShadowOfHeaven-Me: My bad

api/src/main/java/com/github/retrooper/packetevents/protocol/player/User.java
ShadowOfHeaven-Me: I thought it would remove any confusion from this being a persistent store

api/src/main/java/com/github/retrooper/packetevents/protocol/player/User.java
ShadowOfHeaven-Me: Oh I guess with the Typed one a generic can be used

api/src/main/java/com/github/retrooper/packetevents/protocol/player/storage/FastUserStorage.java
ShadowOfHeaven-Me: On any data access, be it write or read, if the class that creates the identifier is initialized later on the initial data size won't be enough. The grow happens only when we're implicitly told that this has happened. It's also safe to do this, since the id is provided internally only

api/src/main/java/com/github/retrooper/packetevents/protocol/player/storage/FastUserStorage.java
ShadowOfHeaven-Me: > if the keys here are plugin instances, this should definetely be a `WeakHashMap` otherwhise this will cause memory leaks when a plugin gets disabled or unloaded
> 
> because of this you will need to do manual synchronization, see some coment below for mor info

Guava's MapMaker is the best solution then

ShadowOfHeaven-Me: > im also not sure if its a good idea to key these ids by "plugin" instance - in my opinion it makes more sense to key the id map by the class of the identifier type, so you can create multiple keys per plugin

The main idea behind this is that you create one key, which holds one custom object, that can encapsulate all the other values associated. Otherwise this could lead to memory being wasted

api/src/main/java/com/github/retrooper/packetevents/protocol/player/storage/StorageValueId.java
ShadowOfHeaven-Me: Can be useful for example if a developer wants to store different types of data throught different parts of joining, where the automatic casting would be only an obstacle

api/src/main/java/com/github/retrooper/packetevents/protocol/player/storage/FastUserStorage.java
ShadowOfHeaven-Me: Fixed it, probably

ShadowOfHeaven-Me:
This won't be very useful unless only 1 instance of PE is used. Channel attributes are more useful for that.
Even tho this storage system is about ~15x times faster compared to any Map after JIT optimizations, it still shouldn't be a bottleneck.
Closed - unnecessary

PR #909, opened 2024-08-06
Add build timestamp to PEVersions class

booky10:
Asked for in [discord](https://canary.discord.com/channels/721686193061888071/721686193515003966/1270216801778925619) and easy to implement

### DIFF
--- a/buildSrc/src/main/kotlin/com/github/retrooper/version/PEVersionTask.kt
+++ b/buildSrc/src/main/kotlin/com/github/retrooper/version/PEVersionTask.kt
@@ -41,9 +41,12 @@ abstract class PEVersionTask : DefaultTask() {
              */
             package $packageName;
             
+            import java.time.Instant;
+            
             public final class PEVersions {
             
                 public static final String RAW = "$version";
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/kotlin/com/github/retrooper/version/PEVersionTask.kt
```diff
@@ -41,9 +41,12 @@ abstract class PEVersionTask : DefaultTask() {
              */
             package $packageName;
             
+            import java.time.Instant;
+            
             public final class PEVersions {
             
                 public static final String RAW = "$version";
+                public static final Instant BUILD_TIMESTAMP = Instant.ofEpochMilli(${System.currentTimeMillis()}L);
```
retrooper, line 49: Where is the BUILD_TIMESTAMP variable used?

booky10: nowhere, it can be used by other plugins to check how old this packetevents build is

PR #912, opened 2024-08-08
Add 1.21.1 version support

booky10:
Nothing changed on the network, 1.21.1 only fixed some server-related crashes related to command argument parsing and block entities


retrooper:
Thanks for the PR.

PR #914, opened 2024-08-09
Refactor NAMED_SOUND_EFFECT to CUSTOM_SOUND_EFFECT as SOUND_EFFECT is doing the work of NAMED_SOUND_EFFECT

libraryaddict:
It'd be nice if we could rename `SOUND_EFFECT` and the wrapper to NAMED_SOUND_EFFECT, but that'd break existing usages.


retrooper:
> It'd be nice if we could rename `SOUND_EFFECT` and the wrapper to NAMED_SOUND_EFFECT, but that'd break existing usages.

We can deprecate a wrapper and create a new one perhaps?

PR #927, opened 2024-08-11
Fix sound exception on 1.21

AoElite:


rafi67000:
LGTM, I tested it using 1.21 and 1.8.9 clients. Works great on both:)

api/src/main/java/com/github/retrooper/packetevents/protocol/item/jukebox/IJukeboxSong.java
```diff
@@ -43,12 +43,10 @@ public interface IJukeboxSong extends MappedEntity, CopyableEntity<IJukeboxSong>
 
     static IJukeboxSong decode(NBT nbt, ClientVersion version, @Nullable TypesBuilderData data) {
         NBTCompound compound = (NBTCompound) nbt;
-
-        Sound sound = Sound.decode(compound.getCompoundTagOrThrow("sound_event"), version);
+        Sound sound = Sound.decode(compound.getStringTagOrThrow("sound_event"), version);
```
booky10: should be changed to just `getTagOrThrow`, this can both be a compound or a string

booky10 (changes requested):

booky10 (approved):
thanks :+1:

PR #932, opened 2024-08-12
Fix dimension type mappings

### LINKED ISSUE #929: Dimension type is null on old servers
**Describe the bug**
I am getting multiple reports of an error relating to dimension type being null on 1.8 servers, Grim is using join game to get the dimension type and it works fine on modern servers. Note that I haven't personally tested this.

https://github.com/GrimAnticheat/Grim/issues/1635

**Software brand**
Paper 1.8.8

**Plugins**
Grim

**How To Reproduce**
Try to get dimension type from join game on a 1.8 server I guess?

**Expected behavior**
Dimension type should not be null

**Screenshots**

**Additional context**

booky10:
Fixes https://github.com/retrooper/packetevents/issues/929

### DIFF
--- a/mappings/world/dimension_type_mappings.json
+++ b/mappings/world/dimension_type_mappings.json
@@ -1,19 +1,19 @@
 {
   "V_1_19": {
-    "minecraft:overworld": 0,
-    "minecraft:the_nether": 1,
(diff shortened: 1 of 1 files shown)


retrooper:
Thanks for the patch!

PR #955, opened 2024-08-20
Read item adventure predicate nbt and property matchers as optionals

booky10:
Fixes https://paste.gg/p/anonymous/ccad925b40214caf8806af1876891eb4


retrooper:
Thanks for the PR!

PR #959, opened 2024-08-21
API for checking if packetevents is terminated

AoElite:
Also adds a setting to toggle kicking players on join if terminated


AoElite:
Looks like the build failed, will check tomorrow

retrooper (approved):

PR #990, opened 2024-09-15
Add wrapper for BossBar packet

Elikill58:
This PR add a wrapper for `BOSS_BAR` packet.

I pre-answer few questions:
- I made my own `Color` enum to prevent switch/case to read color. It also help to other developer to only use right colors
- The constructor isn't with all parameters as they are different between each actions
- For the flags, I didn't convert them to something else that byte as it seems to changed multiple times. Also, I'm reading as UnsignedByte but there isn't writeUnsignedByte, I tried writeShort but that was too much.
- In general, I used description available [here](https://wiki.vg/Protocol#Boss_Bar) and tested on my own server

I suggest you to Squash and merge. That's would be perfect.

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerBossBar.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerBossBar.java
@@ -24,6 +24,7 @@ import com.github.retrooper.packetevents.event.PacketSendEvent;
 import com.github.retrooper.packetevents.protocol.color.DyeColor;
 import com.github.retrooper.packetevents.protocol.packettype.PacketType;
 import com.github.retrooper.packetevents.wrapper.PacketWrapper;
+import net.kyori.adventure.bossbar.BossBar;
 import net.kyori.adventure.text.Component;
 
 public class WrapperPlayServerBossBar extends PacketWrapper<WrapperPlayServerBossBar> {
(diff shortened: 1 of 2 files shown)


booky10:
Using adventure's bossbar classes would be a lot nicer to work with

Elikill58:
> Using adventure's bossbar classes would be a lot nicer to work with

That's can be difficult as only one action has informations about the full bossbar, else it's only a part of it. Should I add something that is a partial bossbar, return null when not full or do nothing?

booky10:
I meant using Bossbar.Color, etc.
Not constructing adventure bossbar objects

Elikill58:
Oh yes, it's good idea. I just made changes about this :)

booky10:
could you also parse the flags sent in the packet?

Elikill58:
I'll check. It will require more testing as it's more versions depending

Elikill58:
That's it! :) (Note: please don't forget to squash, it's better)

api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerBossBar.java
```diff
@@ -0,0 +1,216 @@
...
+
+public class WrapperPlayServerBossBar extends PacketWrapper<WrapperPlayServerBossBar> {
+
+    private UUID uuid;
+    private Action action;
+    private Component title;
+    private float health;
+    private BossBar.Color color;
+    private BossBar.Overlay overlay;
+    private List<BossBar.Flag> flags;
```
booky10: this should be a `Set` (or even an `EnumSet`), as flags are not present multiple times
additionally, using a `Set` would be faster here

booky10 (commented):
one small issue left, the rest looks good :+1:

api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerBossBar.java
Elikill58: I just made changes :)

booky10 (approved):
👍

retrooper:
Thanks for your PR.

PR #1004, opened 2024-09-21
Force usage of plugin classloader to lookup classes of other plugins

### LINKED ISSUE #982: ViaVersion incompatibility?
**Describe the bug**
When pinging and joining the server, I get [this](https://paste.helpch.at/otuhosimos.php) error.

**Software brand**
Using Paper 1.20.6-149

**Plugins**
ViaVersion 5.0.3 (tried using snapshots of 5.0.4 and 5.0.2 but none worked either) and my own Plugin that uses PacketEvents.

booky10:
Works around issues caused by some plugins loading packetevents through the library loader, which doesn't know about the classloader of plugins

Fixes https://github.com/retrooper/packetevents/issues/982


retrooper:
Thank you so much for the PR.

PR #1006, opened 2024-09-21
Fix errors with viaversion and map decorations

### LINKED ISSUE #988: 1.20.4 support
if you try take a map from a boat chest (connected in a 1.20.4 version) you will be kicked from the server
image
not happens on 1.20.1

the server is 1.21 using packetevents 1.21 compatible
the server have viaversion / viabackwards 
i am trying do it  on 1.20.4 and it not works
works on other versions

booky10:
Reproducible on a 1.20.6+ server with a 1.20.4- client when packetevents tries to parse a map decoration type
Caused by packetevents not seeing "`red_x`" as the same as "minecraft:red_x"

This now implements a more generic solution, normalizing all string inputs for registries

Fixes #988


retrooper:
Thank you booky

PR #1014, opened 2024-09-28
fix legacy rail mappings

### LINKED ISSUE #1013: Powered Rail States are Modeled Incorrectly on 1.8 Servers
**Describe the bug**
The blockstates of golden/powered rails when they are in the on/powered state are modeled incorrectly. Ascending golden/powered rails that are activated/on/powered always have their shape returned as "NORTH_SOUTH" regardless of the direction they face.

In addition, east-west rails are modeled as north-east, north-south rails are modeled as north-west, etc... all obviously impossible states.

**Software brand**
PandaSpigot 1.8.8,

**Plugins**
GrimAC, LuckPerms, ViaVeresion

**How To Reproduce**
1. Join with the 1.8.9 vanilla Minecraft client.
2. Place powered rail down and power it. 
3. Get it's blockstate. If you don't have an easy means of doing this you can test my new check being added to GrimAC https://github.com/Axionize/Grim/tree/los-containers. 
Once you've built and installed the plugin, run `/grim verbose` and breakpoint
```
RAILS((player, item, version, data, x, y, z) -> {
----->        switch (data.getShape()) {
```
in `HitBox.java`
then right click on a rail to see the wrong blockstate being returned.

**Expected behavior**
Should return the correct blockstate. I assume you're extracting blockstates for rails with some bit manipulation, and you probably forgot to account for powered rails when they're powered on properly.

**Screenshots**

Screenshot_20240928_180929

ManInMyVan:
should fix #1013

### DIFF
--- a/mappings/block/legacy_block_mappings.json
+++ b/mappings/block/legacy_block_mappings.json
@@ -900,102 +900,134 @@
   },
   "powered_rail": {
     "27:0": {
-      "shape": "north_south"
+      "shape": "north_south",
+      "powered": false
     },
     "27:1": {
-      "shape": "east_west"
+      "shape": "east_west",
+      "powered": false
     },
     "27:2": {
-      "shape": "ascending_east"
+      "shape": "ascending_east",
+      "powered": false
     },
     "27:3": {
-      "shape": "ascending_west"
+      "shape": "ascending_west",
+      "powered": false
     },
     "27:4": {
-      "shape": "ascending_north"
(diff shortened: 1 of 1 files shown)


Axionize (approved):
LGTM

booky10 (changes requested):
You broke detector rails, please restore the correct ids for them

booky10 (approved):

booky10:
Thanks for the quick fix!

PR #1026, opened 2024-10-14
Fix server side for fabric

VendoAU:
I ended up making seperate encoder and decoder classes for clientbound and serverbound packets, not sure if that's the correct way to do it but it works.
I also created a `CustomPipelineUtil` class to match the other environments.


Axionize:
This does not work and leads to an NPE at start when init() is called to try and use reflection to get the methods.

I've gone ahead and fixed it in https://github.com/retrooper/packetevents/pull/1027

PR #1028, opened 2024-10-15
[Feat] bStats Sponge Support

Bram1903:
## Overview
This pull request introduces bStats support for the Sponge platform and updates the bStats version for all other modules within PacketEvents. While updating my AntiHealthIndicator plugin to support Sponge (thanks to PacketEvents adding support), I noticed that PacketEvents lacked bStats integration for Sponge and was also using an outdated bStats version in other modules. This PR addresses both issues.

## Changes
- [x] Added bStats support for Sponge.
- [x] Updated bStats version across all modules.
- [x] Switched to shading bStats via Gradle, instead of copying the `Metrics` class into the codebase.
- [x] Refactored Velocity's bStats initialization to use dependency injection, which is in line with recent changes in the bStats velocity class.
- [x] Moved the Sponge Gradle plugin version specification from `build.gradle.kts` to `libs.versions.toml` for better dependency management.
- [x] Adjusted the placement of the plugin started debug message in the Velocity module to occur when the plugin is actually initialized.


bungeecord/src/main/java/io/github/retrooper/packetevents/bstats/Metrics.java
```diff
@@ -28,8 +28,12 @@
  * bStats collects some data for plugin authors.
  * <p>
  * Check out https://bStats.org/ to learn more about bStats!
+ * <p>
+ * @deprecated This class is deprecated and will be removed in future versions.
+ * Please use 'io.github.retrooper.packetevents.bstats.bungeecord' instead.
```
booky10: @retrooper is this api or internal?
If it's internal, this should be rephrased to "Please shade your own bstats copy instead.".

Bram1903: I am pretty sure this is API. Retrooper once told me I could use this:

image

retrooper: We can make an exception for this one. I probably only told you that you can use it. Let's remove the old class.

retrooper: @booky10 @Bram1903

Bram1903: @retrooper, so do you want me to remove the classes I now marked as deprecated?

retrooper: Yeah.

booky10 (approved):
Looks good :+1:

Although there should be a note next release that internal bstats classes were removed

Bram1903:
> Although there should be a note next release that internal bstats classes were removed

Yeah, I agree. I have to update my plugins as well, but I will keep making use of these shaded classes ;-)

Bram1903:
@retrooper, can you review this PR?

retrooper:
Thank you for your PR.

PR #1032, opened 2024-10-20
Fabric module refactor

### LINKED ISSUE #1025: PacketEvents Fabric Client issue
**Describe the bug**
Sending packets to the user on packetevents does not work or disconnects the player

**Software brand**
1.21 Fabric, fabric loader 0.16.6-1.21, fabric-api-0.102.0+1.21, Packetevents 5.0.0

**Plugins**
Just packetevents and fabric-api-0.102.0+1.21

**How To Reproduce**
Create a fabric mod that uses packetevents and attempt to send a packet to the user

I have a packet listener that will listen to a command being used, take the user for the event, and send them a packet. But the packet sent doesn't do anything, and in some cases, it will just disconnect the player with an error (different depending on each packet). Even user.sendMessage() doesn't do anything, but packets like the block change packet will disconnect the player.

There is usually nothing in the logs and when there is it's different per packet.

booky10:
## Breaking changes

- Nearly all packetevents fabric internals have been changed

## Changes

- Splits sources into client and common sources to prevent accidentally using client code
- Sets up a PacketEvents api instance on both the client and server
- Simplifies network pipeline injection and encoder/decoder handling
- Uses proper SLF4J logger instead of System.out
- Uses fabric access wideners for accessing player connection internals instead of reflection
- Fixes https://github.com/retrooper/packetevents/issues/1025

## Todo

- [x] Look at other platforms and search for stuff which could be missing
  - [x] Associate platform player instance with connection channel
- [x] Testing on client and dedicated server
- [ ] ~~Find a way to create packetevents instances for singleplayer worlds (internal server; maybe using events other mods could listen to?)~~
  - Not possible without loading packetevents on an additional, separate classpath or removing usage of the static instance (which is used everywhere...)

---

Closes https://github.com/retrooper/packetevents/pull/1026
Closes https://github.com/retrooper/packetevents/pull/1027


Bram1903:
@booky10 once again a great PR! Looks really good! Now I can add support for Fabric to my AntiHealthIndicator!

Axionize:
I've worked with Booky to ensure this works. Can we please merge this into upstream so I don't have to maintain this upstream fork and have everyone who contributes to my project build packevents just to build my project?

PR #1043, opened 2024-10-28
Fix block state mapping indexes

booky10:
Refactors this logic to no longer hardcode every mapping index for every version
Also adds a few tests for 1.8 states


retrooper:
Thanks for the changes!

PR #1044, opened 2024-10-28
fix: NPC objects to have unique server team names.

### LINKED ISSUE #1040: All NPC objects share same unique scoreboard team name.
**Describe the bug**
All instances of the NPC class share the same scoreboard team name, which, in effect, means all NPC's share the same metadata of that team. (prefix/suffix). 

https://github.com/retrooper/packetevents/blob/d8a44fe8b6143578afd969c1eb192d782377e2b5/api/src/main/java/com/github/retrooper/packetevents/protocol/npc/NPC.java#L439

**Software brand**
Paper version 1.21.1-105-master@b7ab22d (2024-09-27T18:16:58Z) (Implementing API version 1.21.1-R0.1-SNAPSHOT)

**Expected behavior**
Instances of NPCs should have unique and independent team names.

**Screenshots**
Screenshot 2024-10-24 at 9 10 45 pm
Screenshot 2024-10-24 at 9 10 51 pm
Screenshot 2024-10-24 at 9 11 12 pm

ShreyasAyyengar:
Fixes #1040

@retrooper!

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/protocol/npc/NPC.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/protocol/npc/NPC.java
@@ -40,6 +40,7 @@ public class NPC {
     private NamedTextColor nameColor;
     private Component prefixName;
     private Component suffixName;
+    private String teamName;
     private int displayPing = 0;
     private Location location = new Location(0.0, 0.0, 0.0, 0.0f, 0.0f);
     private ItemStack mainHand = null;
@@ -60,6 +61,7 @@ public class NPC {
         this.nameColor = nameColor;
         this.prefixName = prefixName;
         this.suffixName = suffixName;
+        this.teamName = "custom_name_team-" + id;
     }
 
     public NPC(UserProfile profile, int entityId, @Nullable Component tabName) {
@@ -273,7 +275,7 @@ public class NPC {
         for (Object channel : channels) {
             //Destroy team
             WrapperPlayServerTeams removeTeam =
-                    new WrapperPlayServerTeams("custom_name_team-" + id,
+                    new WrapperPlayServerTeams(teamName,
                             WrapperPlayServerTeams.TeamMode.REMOVE,
                             Optional.empty());
             PacketEvents.getAPI().getProtocolManager().sendPacket(channel, removeTeam);
@@ -410,6 +412,14 @@ public class NPC {
         this.suffixName = nameSuffix;
     }
 
+    public String getTeamName() {
+        return teamName;
+    }
+
+    public void setTeamName(String teamName) {
+        this.teamName = teamName;
+    }
+
     @Nullable
     public Component getTabName() {
         return tabName;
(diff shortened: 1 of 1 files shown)


booky10:
Could you create a private field (type String) which defaults to this string for the teamname, plus a getter/setter?
That way you can decide wether to share prefixes/suffixes or not on your own, instead of being forced

ShreyasAyyengar:
> Could you create a private field (type String) which defaults to this string for the teamname, plus a getter/setter? That way you can decide wether to share prefixes/suffixes or not on your own, instead of being forced

Done!

booky10 (approved):
:+1:

booky10 (changes requested):
Nevermind, this won't work
Please shorten the `teamName` to fit in the character limit of `16`

ShreyasAyyengar:
> Nevermind, this won't work Please shorten the `teamName` to fit in the character limit of `16`

Do team names have character limits? I thought that was just the NPC's name?

booky10:
The team name limit was removed in 1.18, versions before that have a team name limit of 16

ShreyasAyyengar:
> The team name limit was removed in 1.18, versions before that have a team name limit of 16

I see. I've just committed to this fix; let me know if it's appropriate.

booky10:
This won't solve the problem - the default team name will always consist of at least 18 characters

ShreyasAyyengar:
Would it be appropriate if the team name was simplified to just the entityId? Or perhaps `team-{id}`

booky10:
`npc_<id>` or so, yeah - it doesn't need to be complicated

booky10 (approved):

api/src/main/java/com/github/retrooper/packetevents/protocol/npc/NPC.java
```diff
@@ -410,6 +412,16 @@ public void setSuffixName(@Nullable Component nameSuffix) {
         this.suffixName = nameSuffix;
     }
 
+    public String getTeamName() {
+        return teamName;
+    }
+
+    public void setTeamName(String teamName) {
+        if (PacketEvents.getAPI().getServerManager().getVersion().isOlderThan(ServerVersion.V_1_18)) {
```
retrooper, line 420: Use
```java
serverVersion.isOlderThan(ServerVersion.V_1_18)
``` 
instead of accessing the global server version.

retrooper (changes requested):

api/src/main/java/com/github/retrooper/packetevents/protocol/npc/NPC.java
ShreyasAyyengar: Not quite sure what you mean here. I've written the check similarly to many other version-checking condition checks.

https://github.com/retrooper/packetevents/blob/d8a44fe8b6143578afd969c1eb192d782377e2b5/api/src/main/java/com/github/retrooper/packetevents/protocol/npc/NPC.java#L260

https://github.com/retrooper/packetevents/blob/d8a44fe8b6143578afd969c1eb192d782377e2b5/api/src/main/java/com/github/retrooper/packetevents/protocol/npc/NPC.java#L249

https://github.com/retrooper/packetevents/blob/d8a44fe8b6143578afd969c1eb192d782377e2b5/api/src/main/java/com/github/retrooper/packetevents/protocol/npc/NPC.java#L233

retrooper: There's a variable within PacketWrapper, I'm saying use it.

retrooper:
Thanks for the PR.

PR #1047, opened 2024-11-03
Extend coral plant blocktags

Axionize:
Also add new ALIVE_CORAL_PLANTS and DEAD_CORAL_PLANTS blocktags

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/protocol/world/states/defaulttags/BlockTags.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/protocol/world/states/defaulttags/BlockTags.java
@@ -250,6 +250,14 @@ public class BlockTags {
      * Unofficial tag for all glass panes
      */
     public static final BlockTags GLASS_PANES = bind("glass_panes");
+    /**
+     * Unofficial tag for all coral plants
+     */
+    public static final BlockTags ALL_CORAL_PLANTS = bind("alive_coral_plants");
+    /**
+     * Unofficial tag for all dead coral plants
+     */
+    public static final BlockTags DEAD_CORAL_PLANTS = bind("dead_coral_plants");
 
     static {
         BlockTags.WOOL.add(StateTypes.WHITE_WOOL, StateTypes.ORANGE_WOOL, StateTypes.MAGENTA_WOOL, StateTypes.LIGHT_BLUE_WOOL, StateTypes.YELLOW_WOOL, StateTypes.LIME_WOOL, StateTypes.PINK_WOOL, StateTypes.GRAY_WOOL, StateTypes.LIGHT_GRAY_WOOL, StateTypes.CYAN_WOOL, StateTypes.PURPLE_WOOL, StateTypes.BLUE_WOOL, StateTypes.BROWN_WOOL, StateTypes.GREEN_WOOL, StateTypes.RED_WOOL, StateTypes.BLACK_WOOL);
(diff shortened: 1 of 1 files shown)


booky10:
These tags don't exist in vanilla, not even in snapshots, why should we add them?

Axionize:
I added the dead states because of falses in Grim due to dead coral not being modeled as coral with the same hitboxes.

I figured someone somewhere would still want the old behavior of only getting alive coral and dead coral. The important thing is to add the dead coral to the corals blocktag.

I wasn't aware blocktags were just ports off a vanilla feature. I'll delete the new tags and just leave this PR to add the dead corals to the coral tags if you're insistent on it.

booky10:
you can still add them, just move them to the bottom of the class - packetevents already has some "inofficial"/"legacy" block tags

everything above the comments is auto-generated and will be replaced with the next version update

i'm also planning on re-designing the item/block tags, as they no longer represent vanilla's tags properly

Axionize:
I'm not sure exactly where you mean by bottom of the class. At the end of the entries or at the actual bottom below the code? what about adding entries to the blocktag?

booky10:
See how GLASS_BLOCKS or GLASS_PANES are defined

Axionize:
Fixed @booky10

api/src/main/java/com/github/retrooper/packetevents/protocol/world/states/defaulttags/BlockTags.java
```diff
@@ -311,7 +319,7 @@ public class BlockTags {
...
         BlockTags.CORAL_BLOCKS.add(StateTypes.TUBE_CORAL_BLOCK, StateTypes.BRAIN_CORAL_BLOCK, StateTypes.BUBBLE_CORAL_BLOCK, StateTypes.FIRE_CORAL_BLOCK, StateTypes.HORN_CORAL_BLOCK);
         BlockTags.WALL_CORALS.add(StateTypes.TUBE_CORAL_WALL_FAN, StateTypes.BRAIN_CORAL_WALL_FAN, StateTypes.BUBBLE_CORAL_WALL_FAN, StateTypes.FIRE_CORAL_WALL_FAN, StateTypes.HORN_CORAL_WALL_FAN);
-        BlockTags.CORAL_PLANTS.add(StateTypes.TUBE_CORAL, StateTypes.BRAIN_CORAL, StateTypes.BUBBLE_CORAL, StateTypes.FIRE_CORAL, StateTypes.HORN_CORAL);
+        BlockTags.CORAL_PLANTS.add(StateTypes.TUBE_CORAL, StateTypes.BRAIN_CORAL, StateTypes.BUBBLE_CORAL, StateTypes.FIRE_CORAL, StateTypes.HORN_CORAL, StateTypes.DEAD_TUBE_CORAL, StateTypes.DEAD_BRAIN_CORAL, StateTypes.DEAD_BUBBLE_CORAL, StateTypes.DEAD_FIRE_CORAL, StateTypes.DEAD_HORN_CORAL);
```
booky10: The block tags should represent vanilla tags
If you want a tag which contains both alive and dead coral, please add an unofficial tag

booky10 (changes requested):

api/src/main/java/com/github/retrooper/packetevents/protocol/world/states/defaulttags/BlockTags.java
Axionize: should I call it `ALIVE_DEAD_CORAL_PLANTS` `ALIVE_AND_DEAD_CORAL_PLANTS` or `ALL_CORAL_PLANTS` ?

booky10: `ALL_CORAL_PLANTS` sounds more future-proof

Axionize: Done.

booky10 (approved):

PR #1053, opened 2024-11-04
Fix legacy mappings for grass and tallgrass variants

### LINKED ISSUE #1052: Incorrect Legacy Block Mappings for Short and Tall Grass
To start off with some definitions. Short grass is grass that looks like this.

Tall grass is grass that looks like this.

StateTypes.TALL_GRASS is correctly returned for Tall Grass on 1.21 and StateTypes.SHORT_GRASS is correctly returned for Short Grass on 1.21.

The StateTypes.TALL_GRASS is returned for Short Grass on 1.8 and StateTypes.PEONY is returned for Tall Grass on 1.8 Is this intentional? Short grass is called "tallgrass" in 1.8 so does this inconsistency exist to accommodate legacy Minecraft's actual block names? If so, what's the intended way to deal with these blocks consistently across versions?

Axionize:


Axionize:
Fixes https://github.com/retrooper/packetevents/issues/1052

Axionize:
@booky10

booky10:
There is a syntax error in your mappings, see build logs

Axionize:
@booky10 fixed

Axionize:
An upstream issue in grim hinges on this being fixed. I'd appreciate this being done as soon as you have time.

booky10:
I'll hopefully get this done later today, I still want to do some testing before merging to ensure this works properly

Axionize:
Items are similar broken/incorrect just like the blocks. An additional hurdle is that item mappings also don't appear to use blockstates? (or I guess itemstates?).

For example 175:2 corresponds to tallgrass and 175:3 corresponds to large ferns. Every 175 variant is considered a large fern. I don't really care since I only care about the hitboxes of these blocks (and they're all the same) but someone else down the line might. 

Can we make item mappings work with data states?

booky10 (approved):
Still not fully correct, e.g. the upper half of peonies is now "tall_grass[half=upper]" instead of "peony[half=upper]"
But I don't think thats possible to fix with legacy block states

booky10:
Took a bit longer than expected, but its merged now

PR #1062, opened 2024-11-19
Remove amount check from isEmpty

Beaness:
This allows negative item sizes to be sent through packetevents
At the moment ItemStacks with a negative or zero amount can not be serialized through packetevents.
This breaks compatibility for plugins expecting isEmpty to return true if the amount is zero or below

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/protocol/item/ItemStack.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/protocol/item/ItemStack.java
@@ -517,7 +517,7 @@ public class ItemStack {
     }
 
     public boolean isEmpty() {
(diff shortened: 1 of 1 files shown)


booky10:
>  At the moment ItemStacks with a negative or zero amount can not be serialized through packetevents.

This is intentional, this would break with Minecraft 1.20.5+

Vanilla also marks items as empty, if the count is <= 0 (even for versions before 1.20.5)

Beaness:
The vanilla 1.8.9 protocol allows negative itemstacks to be serialized but packetevents replaces this with air due to empty being true on items with amount <= 0

Should there be made another check instead of isEmpty for legacy versions to allow serializing negative items?

booky10:
Yes, that would be a more sensible solution to your specific problem

PR #1070, opened 2024-11-27
Velocity's varint reading/writing optimizations

ShadowOfHeaven-Me:
Basically it. Not sure why I can't disregard my the previous pr changes, but whatever, the current origin should have the exact same code

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/netty/buffer/ByteBufHelper.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/netty/buffer/ByteBufHelper.java
@@ -244,6 +244,7 @@ public class ByteBufHelper {
         return PacketEvents.getAPI().getNettyManager().getByteBufOperator().getIntLE(buffer, readerIndex);
     }
 
+    //Src: https://github.com/jonesdevelopment/sonar/blob/main/common/src/main/java/xyz/jonesdev/sonar/common/fallback/netty/FallbackVarInt21FrameDecoder.java#L69
     public static int readVarInt(Object buf) {
         if (readableBytes(buf) < 4) return readVarIntSmallBuffer(buf);
 
@@ -292,6 +293,7 @@ public class ByteBufHelper {
         return result | (tmp & 0x7F) << 14;
     }
 
+    //Src: https://steinborn.me/posts/performance/how-fast-can-you-write-a-varint/
     public static void writeVarInt(Object buf, int i) {
         if ((i & (0xFFFFFFFF << 7)) == 0) {
             // 1 byte case
--- a/api/src/main/java/com/github/retrooper/packetevents/netty/buffer/ByteBufHelper.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/netty/buffer/ByteBufHelper.java
@@ -19,11 +19,11 @@
 package com.github.retrooper.packetevents.netty.buffer;
 
 import com.github.retrooper.packetevents.PacketEvents;
-import com.github.retrooper.packetevents.protocol.player.ClientVersion;
 
 import java.nio.charset.Charset;
 
 public class ByteBufHelper {
+
     public static int capacity(Object buffer) {
         return PacketEvents.getAPI().getNettyManager().getByteBufOperator().capacity(buffer);
     }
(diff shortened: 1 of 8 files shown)


api/src/main/java/com/github/retrooper/packetevents/netty/buffer/ByteBufHelper.java
```diff
@@ -239,29 +240,82 @@ public static Object resetWriterIndex(Object buffer) {
...
-                writeByte(buffer, value);
-                return;
-            }
-            writeByte(buffer, (value & 0x7F) | 0x80);
-            value >>>= 7;
+    public static int getIntLE(Object buffer, int readerIndex) {
+        return PacketEvents.getAPI().getNettyManager().getByteBufOperator().getIntLE(buffer, readerIndex);
+    }
+
+    //Src: https://github.com/jonesdevelopment/sonar/blob/main/common/src/main/java/xyz/jonesdev/sonar/common/fallback/netty/FallbackVarInt21FrameDecoder.java#L69
```
booky10: this is definetely not the source of this code
either [directly look at velocity](https://github.com/PaperMC/Velocity/blob/9cfcfcf2ed5712e792114a3ab824670e25e23526/proxy/src/main/java/com/velocitypowered/proxy/protocol/netty/MinecraftVarintFrameDecoder.java#L82), or see commit message: https://github.com/jonesdevelopment/sonar/commit/26c31ceeaffdc3c1d3235bd20af18db78c042553
also, please at least mention the license of this piece of code, even if it is GPL

booky10 (changes requested):
Additionally, it may also make sense to move this varint read/write implementation to the ByteBufOperator implementations to prevent way too many getter calls

api/src/main/java/com/github/retrooper/packetevents/netty/buffer/ByteBufHelper.java
ShadowOfHeaven-Me: Ah sorry, didn't see that

ShadowOfHeaven-Me:
...why did this commit here

> Nevermind that, fixed it

retrooper:
It appears that this PR has some conflicts that are unresolved. Do we still intend on integrating this?

netty-common/src/main/java/io/github/retrooper/packetevents/impl/netty/buffer/FastNettyUtils.java
```diff
@@ -0,0 +1,91 @@
...
+        final int bitsToKeep = Integer.numberOfTrailingZeros(atStop) + 1;
+        buf.skipBytes(bitsToKeep >> 3);
+
+        // https://github.com/netty/netty/pull/14050#issuecomment-2107750734
+        int preservedBytes = wholeOrMore & (atStop ^ (atStop - 1));
+
+        // https://github.com/netty/netty/pull/14050#discussion_r1597896639
+        preservedBytes = (preservedBytes & 0x007F007F) | ((preservedBytes & 0x00007F00) >> 1);
+        preservedBytes = (preservedBytes & 0x00003FFF) | ((preservedBytes & 0x3FFF0000) >> 2);
+        return preservedBytes;
```
booky10, line 26: This only reads a maximum of bytes, so only 28 bits - a varint is a 32-bit int, so we are missing one byte
And this whole logic was only designed for a maximum of 21 bits in mind, as this is used in velocity's frame decoder

booky10: We are probably just fine with using the normal way of reading a varint and just looking the first byte beforehand
Most common cases are just one-byte varints anyway

api/src/main/java/com/github/retrooper/packetevents/netty/buffer/ByteBufHelper.java
```diff
@@ -239,29 +240,37 @@ public static Object resetWriterIndex(Object buffer) {
...
+    public static int getIntLE(Object buffer, int readerIndex) {
+        return PacketEvents.getAPI().getNettyManager().getByteBufOperator().getIntLE(buffer, readerIndex);
+    }
+
+    public static int readVarInt(Object buf) {
+        return PacketEvents.getAPI().getNettyManager().getByteBufOperator().readVarInt(buf);
+    }
+
+    //Src: https://steinborn.me/posts/performance/how-fast-can-you-write-a-varint/
+    public static void writeVarInt(Object buf, int i) {
```
booky10, line 252: Any reason why this isn't in `FastNettyUtils`?

booky10 (changes requested):

booky10:
@ShadowOfHeaven-Me do you still plan to further work on this?

netty-common/src/main/java/io/github/retrooper/packetevents/impl/netty/buffer/FastNettyUtils.java
ShadowOfHeaven-Me: I guess

ShadowOfHeaven-Me:
Nah, no need.
Probably the default method is better, due to better inlining and handling the most common case much faster

PR #1074, opened 2024-11-30
Fix velocity precision errors

Beaness:
Having re-encode on by default causes packetevents to write a different velocity due to precision errors.
This uses the raw values for writing to not have any precision loss.

An example of a velocity packet that gets rewritten to a different velocity
```py
PacketEvents read -> X: 1.07625, Y: 0.5035, Z: 0.221
1.8.9 client read -> X: 1.07625 Y: 0.503375 Z: 0.221
```


booky10:
do you have a specific example where packetevents actually changes the value which the client reads?
minecraft encodes velocity values as a short on the protocol level, so even without packetevents there will be a small loss of precision

Beaness:
Yes I have an example in my first message of this pull request. I wrote a small java file to test all 65535 short values and 744/65535 of those values result into a different velocity value on the client. The precision loss for each of those value is ~0.000125

Beaness:
> minecraft encodes velocity values as a short on the protocol level, so even without packetevents there will be a small loss of precision

The problem is the re-encoding of the velocity which does not convert to the same short values back which causes the client to read a different velocity value.

PR #1076, opened 2024-12-03
fix NPE for nbtInterpret

mfnalex:
Fixes a null pointer exception when the "interpret" flag is not set.

I came across this using Items with NBT Compound lores: https://paste.md-5.net/acurekujeq.cs


booky10 (approved):
Thanks for the fix!

PR #1081, opened 2024-12-04
bugfix: Added input validation for window click

Tecnio:
The window click input was not being checked to be in a valid range, solved with UNKNOWN input type.

 There seems to be other Objects which attempts to mitigate this by moduloing the result however this prevents anti-cheat softwares ability to handle these scenarios.


retrooper:
Thanks for the PR!

PR #1096, opened 2024-12-22
Associate bukkit player with user object earlier

### LINKED ISSUE #1036: Player was null on declare commands packet
**Describe the bug**
DeclareCommands packet is sometimes null

**Software brand**
Pufferfish 1.17.1

**Plugins**
Pl-Hide-Pro

**How To Reproduce**
1. Join the server.
2. Error: [Pl-Hide-Pro] Player was null on declare commands packet

**Screenshots**

Screenshot_1
Screenshot_2
Screenshot_3

booky10:
I am not sure if this causes any issues, so I opened a PR before

@retrooper would be nice if you could take a look

Fixes https://github.com/retrooper/packetevents/issues/1036


Axionize:
+1 for this, would allow Grim do some minor optimizations in getting rid of some branching code and *might* fix some of the falses that occur immediately after a player joins

retrooper:
Fixed in latest commit.

PR #1097, opened 2024-12-22
Use correct scheduler for kicking player

HaHaWTH:
In some rare cases, player may be teleported during join and kick will fail,
and all entity operations should be run on EntityScheduler like what they say in [Folia Javadoc](https://jd.papermc.io/folia/1.20/org/bukkit/Bukkit.html#getRegionScheduler())


retrooper:
Thanks for the changes.

PR #1108, opened 2025-01-08
Fix pale oak materials

### LINKED ISSUE #1920: "Pale" blocks cause falses
### Describe the false positive and how to replicate it

The player triggers simulation false positive while trying to step on a pale oak fence.

image

### Grim Log

https://paste.grim.ac/eyuj81e

### Server version

Paper  1.21.4-15-master@bb76819

### Client version

1.21.4

### Plugins

CoreProtect, Essentials, GrimAC, ProtocolLib

ghost:
Fixes https://github.com/GrimAnticheat/Grim/issues/1920

Not sure what happened here, a lot of the pale stuff is incorrect, and it was causing issues with grim anti-cheat because of the bad data, perhaps the pale blocks were copied from an early mc snapshot version and forgot to be updated when mojang eventually fixed them? @booky10 

These fixes were quite tedious, perhaps we should use a python script or some kind of generator to extract this data from mc source files for us automatically, would remove the chance of typos also.

Also, it seems the semantics for isBlocking() and isSolid() is the wrong way around/inconsistent? isBlocking() *should* mean if it has any collision elements and isSolid() means whether it can be built upon, well at least that is how it is in bukkit.


Kadeluxe:
What's wrong here? Isn't it a big issue? Grim anticheat doesn't work properly on 1.21.4 because of that.

ghost:
> What's wrong here? Isn't it a big issue? Grim anticheat doesn't work properly on 1.21.4 because of that.

just compile packetevents yourself with my diff and publish to maven local then compile grim with it

Kadeluxe:
> > What's wrong here? Isn't it a big issue? Grim anticheat doesn't work properly on 1.21.4 because of that.
> 
> just compile packetevents yourself with my diff and publish to maven local then compile grim with it

I do that, just wondering what's possibly wrong here

booky10 (approved):
There is no proper definition of what "blocking" or "solid" means for packetevents, but these changes align with what is set for oak materials :+1:

PR #1109, opened 2025-01-08
Generator idea

ghost:
Generate (in this case StateTypes) from the underlying minecraft data, prevents mistakes and allows easy updating.
It also lets us expand StateTypes with ease to include other minecraft data like piston push reactions and map colors by just changing a couple lines of code.

Also fixes the problem of https://github.com/retrooper/packetevents/pull/1108
and issues like this:
https://github.com/retrooper/packetevents/commit/a4a375f41181d38771f12ed6116be1513dae132f
https://github.com/retrooper/packetevents/commit/2440b649812a4c705af9b71b3d8c11f463fcfd5a
https://github.com/retrooper/packetevents/commit/25dfb3c5b9a1fcec5b8a762725f0b4ea0744ece0

Here is the current output of the generator from the code: https://pastebin.com/nV6zDzUi
The MaterialType doesn't quite match since the MaterialType enum seems to be very arbitrary, but I will make it match 1:1 if you are interested in the generator idea.

Also the order of current StateTypes is slightly different compared to in minecraft so the generated output entries after a point have a slightly different order, you can run both through https://dedupelist.com/ and sort the list to see the diff easier.

### DIFF
--- a/api/build.gradle.kts
+++ b/api/build.gradle.kts
@@ -39,6 +39,11 @@ dependencies {
     testImplementation(testlibs.slf4j)
     testImplementation(testlibs.bundles.junit)
     testImplementation(libs.netty)
+
+    // Minecraft jar input for generator
+    //compileOnly(files("C:/Users/User/Documents/newwind_source/newwind_plugins/NWNMS/libs/paper-1.21.4.jar"))
+    //compileOnly("com.mojang:datafixerupper:8.0.16")
+    //compileOnly("io.papermc.paper:paper-api:1.21.4-R0.1-SNAPSHOT")
 }
 
 mappingCompression {
--- a/api/src/main/java/com/github/retrooper/packetevents/generator/StateTypesGenerator.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/generator/StateTypesGenerator.java
@@ -21,7 +21,7 @@ import java.util.Locale;
 public final class StateTypesGenerator {
 
     public static void generateStateTypesString() throws Exception {
-        System.out.println("Generating state types string:");
+        System.out.println("Generating state types...");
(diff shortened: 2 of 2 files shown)


booky10:
https://github.com/booky10/PacketEventsGenerators/blob/master/src%2Fmain%2Fjava%2Fdev%2Fbooky%2Fgeneration%2Fgenerators%2FStateTypesGenerator.java :eyes:

ghost:
> https://github.com/booky10/PacketEventsGenerators/blob/master/src%2Fmain%2Fjava%2Fdev%2Fbooky%2Fgeneration%2Fgenerators%2FStateTypesGenerator.java 👀

Didn't know about this, where is this generator getting its data from? If it was used for the pale blocks then there is definitely some kind of bug in it, as almost all of them were incorrect in some way, also, notably some fields such as isShapeExceedsCube are missing from that generator

booky10:
> Didn't know about this, where is this generator getting its data from?

From the vanilla sources, see the gradle buildscript

> notably some fields such as isShapeExceedsCube are missing from that generator

`isShapeExceedsCube` was indeed missing (fixed in https://github.com/booky10/PacketEventsGenerators/commit/13cf9f6688b1959dc363d53ccbc895992fb87b5b) - which other "fields" do you mean?

> If it was used for the pale blocks then there is definitely some kind of bug in it, as almost all of them were incorrect in some way

Both blocking and solid (which you changed in #1108) are based on the same code in this pr and my generator - so how are you trying to solve this issue?

---

The problem currently is that there is no "official" definition (for packetevents) which blocks count as "solid" and which blocks count as "blocking"

Vanilla's "solid" is some legacy value which is determined by hard-coded flags or calculated based on the size of the static collision shape (as long as the collision shape isn't dynamic)

PR #1134, opened 2025-02-10
Make PE painting IDs consistent

Axionize:
Minecraft sends paintings over the network with ids not representing registration order, but rather alphabetical order.

This commit reorders the painting mappings to all be done in alphabetical order so the index IDs match up with vanilla.

Note that since our registry starts counting from 0 instead of 1, values obtained from listening to networking packets require developers to subtract one so registry entries match up.


booky10 (changes requested):
Please revert your changes to the registry order for 1.19 and 1.7.10, this is incorrect

Minecraft uses dynamic registry loading for painting variants starting with 1.21 which causes the painting on a vanilla server to be sorted alphabetically
(There is the possibility of custom painting variants because of dynamic loading and you shouldn't rely on these static registry values anyway)

booky10 (approved):

Axionize:
> Please revert your changes to the registry order for 1.19 and 1.7.10, this is incorrect

Thanks for the fast response this time, done, I changed 1.19 and 1.7.10 ones immediately, that was a mistake
 
> Minecraft uses dynamic registry loading for painting variants starting with 1.21 which causes the painting on a vanilla server to be sorted alphabetically (There is the possibility of custom painting variants because of dynamic loading and you shouldn't rely on these static registry values anyway)

Dynamic loading will only happen for modded environments and in that case we parse Registry_Data for the `minecraft:painting_variant` registreyKey, but this makes it significantly easier for non-modded environments to deal with and will match expectations

PR #1135, opened 2025-02-10
Add new methods to Direction Enum and optimize by caching values array

Axionize:


api/src/main/java/com/github/retrooper/packetevents/protocol/world/Direction.java
```diff
@@ -18,28 +18,45 @@
...
+        this.axis = axis;
+        this.vec3i = vec3i;
     }
 
     public int getHorizontalIndex() {
         return horizontalIndex;
     }
 
     private static final Direction[] HORIZONTAL_VALUES = {NORTH, SOUTH, WEST, EAST};
+    private static final Direction[] VALUES = values(); // Cache the values array
```
booky10, line 49: Any reason for this field? Its not used anywhere in your code

Axionize: A mistake I made in my haste to get all the changes I needed made. Should've been public.

Is a common trick used for converting integers to an enum by their ordinals when `values()` doesn't change. Instead of making an expensive `Direction.values()[index]` call every time we want to convert int/Integer -> Direction or other enums we can instead do the cheap array lookup of `Direction.VALUES[index]` instead only eating the cost of values() on init.

booky10: The problem is that arrays in java are mutable (as long as https://openjdk.org/jeps/8261007 isnt released)
So an array like this could be accidentally modified and break every plugin relying on this

booky10 (approved):

PR #1137, opened 2025-02-12
Add clone() to Vector classes and add new Mutable Vector Classes

Axionize:
Allow Vector3d, f, and i classes to be cloned and add setX, setY, and setZ methods to match Bukkit Vector abilities.

This is intended as part of the work in porting Grim to being multi-platform, which involves removing all usage of Bukkit Vectors, nevertheless we need to replace their functionality and would like to remain compatible with upstream PE.

I can understand if mutability in Vector3d, f, and i are things that you don't want changed, and if that's the case I'll instead work with you in finding a solution that allows us to not break from using PE vectors, such as  adding new Vector3dI, Vector3fI, etc.. classes as immutable variants and having Vector3d and Vector3dI share the same interface which is used internally in PE code that takes the vectors as arguments instead of `Vector3d`. 

This would allow developers to choose immutability if they so wish and use them with the PE methods that take Vector objects as arguments (there aren't that many so it'll be an easy fix).

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/util/Vector3d.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/util/Vector3d.java
@@ -19,14 +19,12 @@
 package com.github.retrooper.packetevents.util;
 
 import com.github.retrooper.packetevents.protocol.nbt.NBT;
-import com.github.retrooper.packetevents.protocol.nbt.NBTDouble;
 import com.github.retrooper.packetevents.protocol.nbt.NBTList;
 import com.github.retrooper.packetevents.protocol.nbt.NBTNumber;
-import com.github.retrooper.packetevents.protocol.nbt.NBTType;
 import com.github.retrooper.packetevents.protocol.player.ClientVersion;
 import com.github.retrooper.packetevents.protocol.world.BlockFace;
-import com.github.retrooper.packetevents.wrapper.PacketWrapper;
 
+import com.github.retrooper.packetevents.wrapper.PacketWrapper;
 import java.util.Objects;
 import org.jetbrains.annotations.NotNull;
 
@@ -107,35 +105,6 @@ public class Vector3d implements VectorInterface3d {
         }
     }
 
-    public static Vector3d read(PacketWrapper<?> wrapper) {
-        double x = wrapper.readDouble();
-        double y = wrapper.readDouble();
-        double z = wrapper.readDouble();
-        return new Vector3d(x, y, z);
-    }
-
-    public static void write(PacketWrapper<?> wrapper, Vector3d vector) {
-        wrapper.writeDouble(vector.x);
-        wrapper.writeDouble(vector.y);
-        wrapper.writeDouble(vector.z);
-    }
-
-    public static Vector3d decode(NBT tag, ClientVersion version) {
-        NBTList<?> list = (NBTList<?>) tag;
-        double x = ((NBTNumber) list.getTag(0)).getAsDouble();
-        double y = ((NBTNumber) list.getTag(1)).getAsDouble();
-        double z = ((NBTNumber) list.getTag(2)).getAsDouble();
-        return new Vector3d(x, y, z);
-    }
-
-    public static NBT encode(Vector3d vector3d, ClientVersion version) {
-        NBTList<NBTDouble> list = new NBTList<>(NBTType.DOUBLE, 3);
-        list.addTag(new NBTDouble(vector3d.x));
-        list.addTag(new NBTDouble(vector3d.y));
-        list.addTag(new NBTDouble(vector3d.z));
-        return list;
-    }
-
     @Override
     public double getX() {
         return x;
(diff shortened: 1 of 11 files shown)


Axionize:
> Allow Vector3d, f, and i classes to be cloned and add setX, setY, and setZ methods to match the functionality.
> 
> This is intended as part of the work in porting Grim to being multi-platform, which involves removing all usage of Bukkit Vectors, nevertheless we need to replace their functionality and would like to remain compatible with upstream PE.

I can understand if mutability in Vector3d, f, and i are things that you don't want changed, and if that's the case I'll consider adding new Vector3dI, Vector3fI, etc.. classes as immutable variants and having them share the same interface.

booky10 (changes requested):
You would be making immutable vector types mutable, which is a major change in behavior and will have unintended side effects
Please either return a new vector instead of modifying the existing one - or create a new mutable vector type

Axionize:
Understandably, I didn't want to have to do this but I get it.

I'll make `Vector3XMutable` classes for each type

Axionize:
@booky10 I've made appropriate changes I think you would find acceptable. I've added a mutable variant and an interface they both share. Is this design acceptable to you?

api/src/main/java/com/github/retrooper/packetevents/util/Vector3f.java
```diff
@@ -30,7 +31,7 @@
  * @author retrooper
  * @since 1.8
  */
-public class Vector3f {
+public class Vector3f implements Cloneable {
```
booky10: Shouldnt this extend the interface?

api/src/main/java/com/github/retrooper/packetevents/util/Vector3i.java
```diff
@@ -33,7 +34,7 @@
  * @author retrooper
  * @since 1.7
  */
-public class Vector3i {
+public class Vector3i implements Cloneable {
```
booky10: Shouldnt this extend the interface?

api/src/main/java/com/github/retrooper/packetevents/util/VectorMutable3i.java
```diff
@@ -0,0 +1,291 @@
...
+
+    public int getY() {
+        return y;
+    }
+
+    public int getZ() {
+        return z;
+    }
+
+    public boolean equals(VectorMutable3i other) {
```
booky10, line 161: Shouldnt the parameter be the interface? (Same for every other method in this class)

api/src/main/java/com/github/retrooper/packetevents/util/VectorMutable3i.java
```diff
@@ -0,0 +1,291 @@
...
+    public int hashCode() {
+        return Objects.hash(x, y, z);
+    }
+
+    public Vector3d toVector3d() {
+        return new Vector3d(x, y, z);
+    }
+
+    public VectorMutable3i add(int x, int y, int z) {
+        return new VectorMutable3i(this.x + x, this.y + y, this.z + z);
```
booky10: Why does this return a new vector? (same for every other method here)

api/src/main/java/com/github/retrooper/packetevents/util/VectorMutable3f.java
```diff
@@ -0,0 +1,237 @@
...
+
+    public float getY() {
+        return y;
+    }
+
+    public float getZ() {
+        return z;
+    }
+
+    public boolean equals(VectorMutable3f other) {
```
booky10, line 114: Shouldnt the parameter be the interface? (Same for every other method in this class)

api/src/main/java/com/github/retrooper/packetevents/util/VectorMutable3d.java
```diff
@@ -0,0 +1,318 @@
...
+    public double getY() {
+        return y;
+    }
+
+    @Override
+    public double getZ() {
+        return z;
+    }
+
+    public boolean equals(VectorMutable3d other) {
```
booky10, line 122: Shouldnt the parameter be the interface? (Same for every other method in this class)

api/src/main/java/com/github/retrooper/packetevents/util/Vector3i.java
```diff
@@ -158,24 +159,29 @@ public int getZ() {
         return z;
     }
 
+    public boolean equals(Vector3i other) {
```
booky10, line 162: Shouldnt the parameter be the interface?

api/src/main/java/com/github/retrooper/packetevents/util/Vector3f.java
```diff
@@ -111,24 +112,29 @@ public float getZ() {
         return z;
     }
 
+    public boolean equals(Vector3f other) {
```
booky10, line 115: Shouldnt the parameter be the interface?

api/src/main/java/com/github/retrooper/packetevents/util/Vector3d.java
```diff
@@ -135,36 +136,44 @@ public static NBT encode(Vector3d vector3d, ClientVersion version) {
...
     public double getY() {
         return y;
     }
 
+    @Override
     public double getZ() {
         return z;
     }
 
+    public boolean equals(Vector3d other) {
```
booky10, line 123: Shouldnt the parameter be the interface?

api/src/main/java/com/github/retrooper/packetevents/util/VectorInterface3d.java
```diff
@@ -0,0 +1,7 @@
+package com.github.retrooper.packetevents.util;
+
+public interface VectorInterface3d extends Cloneable {
```
booky10: This interface should have more methods which are present in both implementstions

api/src/main/java/com/github/retrooper/packetevents/util/VectorInterface3f.java
```diff
@@ -0,0 +1,7 @@
+package com.github.retrooper.packetevents.util;
+
+public interface VectorInterface3f extends Cloneable {
```
booky10, line 3: This interface should have more methods which are present in both implementstions

api/src/main/java/com/github/retrooper/packetevents/util/VectorInterface3i.java
```diff
@@ -0,0 +1,7 @@
+package com.github.retrooper.packetevents.util;
+
+public interface VectorInterface3i extends Cloneable {
```
booky10, line 3: This interface should have more methods which are present in both implementstions

booky10 (changes requested):
This design looks good 👍 
You could maybe move now duplicate logic to the interface (e.g. read/write/equals/etc.)

api/src/main/java/com/github/retrooper/packetevents/util/VectorMutable3i.java
Axionize: JVM can't devirtualize and we have to use getter methods if we use the interface. I could have separate equals methods for `Vector3d` and `VectorMutable3d` in both `Vector3d` and `VectorMutable3d` I suppose...

The original idea was to have a fast equals path when you know you're working with a Vector3d and another Vector3d or a VectorMutable3d and VectorMutable3d. When you're working purely with the interface though you'll call the slower `equals(Object)`

Axionize:
Ok I pushed some new changes to VectorInterface3d but not the others just as a demo. I don't actually really like this design because there are few times if any where you want to directly work with and between mutable and immutable vectors, they should stay largely separate and I hate having to eat the virtualization cost of the abstraction.

I would appreciate it if you went through VectorInterface3d and told me what methods you would be okay with stripping, because I personally want to keep it very minimal.

api/src/main/java/com/github/retrooper/packetevents/util/VectorMutable3d.java
```diff
@@ -0,0 +1,335 @@
...
+    @Override
+    public VectorMutable3d subtract(VectorInterface3d other) {
+        return subtract(other.getX(), other.getY(), other.getZ());
+    }
+
+    @Override
+    public VectorMutable3d multiply(double x, double y, double z) {
+        this.x -= x;
+        this.y -= y;
+        this.z -= z;
```
Cyramek, line 189:
```java
        this.x *= x;
        this.y *= y;
        this.z *= z;
```

PR #1140, opened 2025-02-12
Improved exception handling

ytnoos:
In my opinion this is the correct way to do it, if you have better ideas let me know. It should be done also for other Decoders but i'm too lazy ^^"

### DIFF
--- a/spigot/src/main/java/io/github/retrooper/packetevents/injector/handlers/PacketEventsDecoder.java
+++ b/spigot/src/main/java/io/github/retrooper/packetevents/injector/handlers/PacketEventsDecoder.java
@@ -27,8 +27,8 @@ import com.github.retrooper.packetevents.util.ExceptionUtil;
 import com.github.retrooper.packetevents.util.PacketEventsImplHelper;
 import com.github.retrooper.packetevents.wrapper.play.server.WrapperPlayServerDisconnect;
 import io.github.retrooper.packetevents.injector.connection.ServerConnectionInitializer;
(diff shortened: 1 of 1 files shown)


spigot/src/main/java/io/github/retrooper/packetevents/injector/handlers/PacketEventsDecoder.java
```diff
@@ -67,36 +79,44 @@ public void decode(ChannelHandlerContext ctx, ByteBuf buffer, List<Object> out)
...
-                } catch (Exception ignored) { // There may (?) be an exception if the player is in the wrong state...
-                    // Do nothing.
-                }
-                ctx.channel().close();
-                if (player != null) {
-                    FoliaScheduler.getEntityScheduler().runDelayed(player, (Plugin) PacketEvents.getAPI().getPlugin(), (o) -> player.kickPlayer("Invalid packet"), null, 1);
-                }
+        boolean debug = PacketEvents.getAPI().getSettings().isDebugEnabled() || SpigotReflectionUtil.isMinecraftServerInstanceDebugging();
+        // We log exceptions only if the server is in debug mode or the player is fully connected to the server.
+        if (debug || (user != null && user.getDecoderState() == ConnectionState.HANDSHAKING)) {
```
booky10: shouldn't this be `user.getDecoderState() != ConnectionState.HANDSHAKING`?

booky10 (changes requested):

booky10 (approved):
this should also be done for the encoder in the future

PR #1146, opened 2025-02-18
Wrong method name

Devlrxxh:


booky10:
This would be a major breaking change and I don't see an issue with the current name
Is there any reason for changing this?

Souipi:
In MCP it's GameProfile so the method's name looks good to me. But "UserProfile" it's just to create a difference ?

booky10:
There is no reason to change this method name, this would be a major breaking change

PR #1147, opened 2025-02-24
Port SpigotConversionUtil to fabric platform

booky10:


Axionize:
You going to provide this for every version? If not, I'd rather just leave this here as reference and have it implemented when needed by end user for what they need.

PR #1149, opened 2025-02-24
Fire UserLoginEvent() on Fabric

Axionize:


fabric/src/main/java/io/github/retrooper/packetevents/mixin/PlayerListMixin.java
```diff
@@ -33,26 +41,69 @@
...
+        if (user == null) {
+            // We did not inject this user
+            Object channelObj = api.getPlayerManager().getChannel(player);
+
+            // Check if it's a fake connection
+            if (!FakeChannelUtil.isFakeChannel(channelObj) &&
+                (!api.isTerminated() || api.getSettings().isKickIfTerminated())) {
+                // Kick the player if they're not a fake player
+                player.connection.disconnect(Component.literal("PacketEvents 2.0 failed to inject"));
+                LOGGER.warn("Failed to inject PacketEvents for player: {}", player.getName().getString());
```
booky10: Is there any reason for this log message?

fabric/src/main/java/io/github/retrooper/packetevents/mixin/PlayerListMixin.java
```diff
@@ -33,26 +41,69 @@
...
+     * @reason Associate connection instance with player instance and handle login event
+     */
+    @Inject(
+        method = "placeNewPlayer",
+        at = @At(
+            value = "INVOKE",
+            target = "Lnet/minecraft/server/players/PlayerList;broadcastAll(Lnet/minecraft/network/protocol/Packet;)V",
+            shift = At.Shift.AFTER
+        )
+    )
```
booky10: This doesnt look like a stable injection point, how about injecting into the head of "addNewPlayer"? (See injection method above)

fabric/src/main/java/io/github/retrooper/packetevents/mixin/PlayerListMixin.java
```diff
@@ -33,26 +41,69 @@
...
+    private void onPlayerLogin(
+        Connection connection, ServerPlayer player,
+        CommonListenerCookie cookie, CallbackInfo ci
+    ) {
+        PacketEventsAPI<?> api = PacketEvents.getAPI();
+        Channel channel = connection.channel;
+
+        // Get or create User instance
```
booky10: What does this comment mean? There will be no user created here

fabric/src/main/java/io/github/retrooper/packetevents/mixin/PlayerListMixin.java
```diff
@@ -33,26 +41,69 @@
...
+                (!api.isTerminated() || api.getSettings().isKickIfTerminated())) {
+                // Kick the player if they're not a fake player
+                player.connection.disconnect(Component.literal("PacketEvents 2.0 failed to inject"));
+                LOGGER.warn("Failed to inject PacketEvents for player: {}", player.getName().getString());
+            }
+            return;
+        }
+
+        // Associate connection with player
+        // api.getInjector().setPlayer(channel, player);
```
booky10: Remove this comment, this doesn't happen here

booky10 (changes requested):
Also, I think your continuation indent is set to 4 instead of 8 and you re-indented the whole file because of that

fabric/src/main/java/io/github/retrooper/packetevents/mixin/PlayerListMixin.java
Axionize: I'm injecting at the same place Paper calls OnPlayerJoin() which you listen to to fire this event on Spigot. I call at the earliest possible point at which the ServerPlayer can be gotten by calling PlayerList methods.

fabric/src/main/java/io/github/retrooper/packetevents/mixin/PlayerListMixin.java
Axionize: replicating Spigot injection behaviour

Axionize:
> Also, I think your continuation indent is set to 4 instead of 8 and you re-indented the whole file because of that

Fixed

booky10:
> > Also, I think your continuation indent is set to 4 instead of 8 and you re-indented the whole file because of that
> 
> Fixed

Are you sure about that? The diff is still there

booky10 (approved):
there are still three lines using a continuation indent of 4, but that doesnt really matter at the moment

PR #1154, opened 2025-03-01
Fix PacketEncoder behaviour and make getTasksAfterSend() work on Fabric

Axionize:
Execute getTasksAfterSend() on fabric and replicate other several other PacketEncoder behaviours from other platforms on Fabric.

### DIFF
--- a/fabric/src/main/java/io/github/retrooper/packetevents/handler/PacketEncoder.java
+++ b/fabric/src/main/java/io/github/retrooper/packetevents/handler/PacketEncoder.java
@@ -18,15 +18,27 @@
 
 package io.github.retrooper.packetevents.handler;
 
+import com.github.retrooper.packetevents.PacketEvents;
+import com.github.retrooper.packetevents.event.PacketSendEvent;
+import com.github.retrooper.packetevents.event.ProtocolPacketEvent;
+import com.github.retrooper.packetevents.exception.CancelPacketException;
+import com.github.retrooper.packetevents.exception.InvalidDisconnectPacketSend;
+import com.github.retrooper.packetevents.exception.PacketProcessException;
+import com.github.retrooper.packetevents.protocol.ConnectionState;
 import com.github.retrooper.packetevents.protocol.PacketSide;
 import com.github.retrooper.packetevents.protocol.player.User;
+import com.github.retrooper.packetevents.util.ExceptionUtil;
 import com.github.retrooper.packetevents.util.PacketEventsImplHelper;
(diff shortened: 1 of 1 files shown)


fabric/src/main/java/io/github/retrooper/packetevents/handler/PacketEncoder.java
```diff
@@ -46,15 +59,100 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
...
+        promise.addListener(p -> this.promise = oldPromise);
+        this.promise = promise;
+
+        // Process the packet and execute post-send tasks (matches Spigot)
+        handlePacket(ctx, in, promise);
+
+        // Check for empty packets last (matches Spigot)
         if (!in.isReadable()) {
             in.release();
+            throw CancelPacketException.INSTANCE;
```
booky10: as this isn't a `MessageToMessageEncoder` (like spigot), just do `return` instead of throwing an exception

fabric/src/main/java/io/github/retrooper/packetevents/handler/PacketEncoder.java
```diff
@@ -46,15 +59,100 @@ public void write(ChannelHandlerContext ctx, Object msg, ChannelPromise promise)
...
         }
+
+        super.exceptionCaught(ctx, cause);
+    }
+
+    // Placeholder for Minecraft server debugging check (Fabric-specific)
+    private boolean isMinecraftServerInstanceDebugging() {
+        // TODO: Implement Fabric-specific debugging check
+        return false;
     }
```
booky10, line 157: just remove this for now

booky10 (changes requested):

fabric/src/main/java/io/github/retrooper/packetevents/handler/PacketEncoder.java
Axionize: Are you certain? Its easier to leave as a placeholder for later, and the TODO is a nice reminder. I don't use this now but structurally if I or someone else were to add support later this would make it much easier.

fabric/src/main/java/io/github/retrooper/packetevents/handler/PacketEncoder.java
Axionize: I'm not certain that's correct but I'll take your word for it. Done.

booky10: which part are you uncertain about?

PR #1165, opened 2025-03-14
Adding ability to add player-specific packet listeners

OOP-778:
My motivate for this PR is simple, I'm too performance obsessed and doing ConcurrentHashMap calls to get data based is not for me, maybe others feel the same way, so the whole point of this PR is to make so you can register player-specific listeners so you can store state within the listener itself.

TODO:

- [x] Separate EventManager impl to BaseEventManager
- [x] Recreate the whole EventManager structure to allow inheriting packet listeners
- [x] Store EventManager in user object itself
- [x] Change the calls to be made for the User EventManager not global.

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/PacketEventsAPI.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/PacketEventsAPI.java
@@ -18,6 +18,7 @@
 
 package com.github.retrooper.packetevents;
 
+import com.github.retrooper.packetevents.event.BaseEventManager;
 import com.github.retrooper.packetevents.event.EventManager;
 import com.github.retrooper.packetevents.injector.ChannelInjector;
 import com.github.retrooper.packetevents.manager.player.PlayerManager;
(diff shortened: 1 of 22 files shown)


booky10 (changes requested):
although your PR isnt finished yet, please don't do any major breaking changes
currently, you would be breaking every single plugin which uses packetevents, as you changed the EventManager to be an interface and not a class

OOP-778:
Would Java cry that the class has been changed from class to Interface? since none of the methods been removed.

OOP-778:
So in theory the signature did not change of it apart that it became an interface

booky10:
> Would Java cry that the class has been changed from class to Interface? since none of the methods been removed.

Yes, it will

OOP-778:
Okay imma make it abstract class

OOP-778:
Gotta change the naming a bit of the event managers butt I think the structure-rework is done and should be backwards compatible.

OOP-778:
@booky10 Pretty much done, apart from cleaning up some parts, lmk what your thoughts on the changes

OOP-778:
P.S: Tested with Grim for backwards-compatability

Beaness:
You could also use a ThreadLocal to store state bound to netty threads, example: https://github.com/Beaness/UnnamedAC/blob/6042a33fa1fb53ca6b49188a53f4c399a8202975/plugin/src/main/java/me/beanes/acid/plugin/player/PlayerManager.java#L10

retrooper:
This unfortunately isn't something that we plan to add to PacketEvents.

PR #1177, opened 2025-03-29
Eliminate usage of static ProtocolManager fields which should be accessed on an instance basis 

Axionize:
Remove usage of direct external/public static references to ProtocolManager.CHANNELS and ProtocolManager.USERS

Fields will remain accessible for backwards compatibility in the unlikely case of a developer meddling these values in their own code.

All usage of both these fields is now wrapped in interface methods which can be overwritten by other ProtocolManagers.


ghost:
Useful tip, you can see if a field is being "used" by using github code search https://github.com/search?q=protocolmanager+%22channels%22+retrooper+language%3AJava+&type=code

it only seems to be used by grim, so if they updated their code it'd be mostly safe to remove.

ManInMyVan:
> Fields will remain accessible for backwards compatibility in the unlikely case of a developer meddling these values in their own code.

it isn't possible to make private fields in interfaces anyways

PR #1180, opened 2025-03-30
Fix Goat Horn Cooldown

### LINKED ISSUE #2053: Holding the Goat Horn while it is on cooldown and attempting to right-click will prevent movement and trigger a NoSlow false alarm.
### Describe the false positive and how to replicate it

1. Holding the Goat_Horn and continuously holding the right button, then moving.
2. NoSlow，NoSlow，NoSlow

### Grim version

https://github.com/GrimAnticheat/Grim/actions/runs/13863014830
It seems to have started appearing from this commit.

### Server version

Paper1.21.4

### Client version

1.21.4

### Plugins

Grimac

Axionize:
Closes https://github.com/GrimAnticheat/Grim/issues/2053. Goat horn cooldown is 7 seconds

Note that we need a way to specify the cooldown group for the future. Right now it doesn't matter because it falls back minecraft:goat_horn in vanilla which is only used for that item.

### DIFF
--- a/mappings/item_base_components/V_1_21_2.json
+++ b/mappings/item_base_components/V_1_21_2.json
@@ -4801,7 +4801,8 @@
     "item_model": "E21pbmVjcmFmdDpnb2F0X2hvcm4=",
     "item_name": "CggACXRyYW5zbGF0ZQAYaXRlbS5taW5lY3JhZnQuZ29hdF9ob3JuAA==",
     "max_stack_size": "AQ==",
-    "rarity": "AQ=="
+    "rarity": "AQ==",
+    "use_cooldown": "QOAAAAETbWluZWNyYWZ0OmdvYXRfaG9ybg=="
   },
   "goat_spawn_egg": {
     "item_model": "GG1pbmVjcmFmdDpnb2F0X3NwYXduX2VnZw==",
--- a/mappings/item_base_components/V_1_21_4.json
+++ b/mappings/item_base_components/V_1_21_4.json
@@ -4835,7 +4835,8 @@
     "item_model": "E21pbmVjcmFmdDpnb2F0X2hvcm4=",
     "item_name": "CggACXRyYW5zbGF0ZQAYaXRlbS5taW5lY3JhZnQuZ29hdF9ob3JuAA==",
     "max_stack_size": "AQ==",
-    "rarity": "AQ=="
+    "rarity": "AQ==",
+    "use_cooldown": "QOAAAAETbWluZWNyYWZ0OmdvYXRfaG9ybg=="
   },
   "goat_spawn_egg": {
     "item_model": "GG1pbmVjcmFmdDpnb2F0X3NwYXduX2VnZw==",


SamB440:
GOAT

baaa

Axionize:
Did I... make a typo?

booky10 (changes requested):
what is this supposed to resolve? the base components are auto-generated for a specific minecraft version and should not be edited manually

Axionize:
The component data for the goat_horn is wrong. The goat_horn (and its variants) have a cooldown of 7.0 seconds on use and all have the default cooldown name.

Unlike chorus fruits however, which explicitly have the cooldown group set as optional.empty(), goat_horn has it set to null which fallsback to the default `minecraft:goat_horn`  (and its registered in a custom location unlike chorus fruits, ender pearls, and wind charges, likely why your automated tools missed it).

This PR originally didn't build properly because I thought you were base64 encoding the cooldown (4 bytes) as a float, I now see you were base64 encoding 
- 4 bytes for seconds in float
- 1 byte (0 for no cooldown group, 1 if there is a group)
- varint for length of string/resource identifier if there is a group
- resource identifier/string is there is a group

```
    public static ItemUseCooldown read(PacketWrapper<?> wrapper) {
        float seconds = wrapper.readFloat();
        ResourceLocation cooldownGroup = wrapper.readOptional(PacketWrapper::readIdentifier);
        return new ItemUseCooldown(seconds, cooldownGroup);
    }
```

For reference this is the python code I used to generate the new `"use_cooldown"` in the JSON I'm about to commit

```
import base64
import struct

def encode_use_cooldown(cooldown_seconds: float, group: str = None) -> str:
    out = bytearray()
    
    # 1. Write float (big endian)
    out += struct.pack('>f', cooldown_seconds)
    
    if group is not None:
        # 2. Write boolean true
        out += b'\x01'
        
        # 3. Encode group string with VarInt-prefixed length
        group_bytes = group.encode('utf-8')
        length = len(group_bytes)
        
        # Minecraft-style VarInt encoding
        while True:
            temp = length & 0x7F
            length >>= 7
            if length != 0:
                temp |= 0x80
            out.append(temp)
            if length == 0:
                break
        
        out += group_bytes
    else:
        # 2. Write boolean false (no group)
        out += b'\x00'
    
    # Return Base64 string
    return base64.b64encode(out).decode('ascii')

print(encode_use_cooldown(7.0, "minecraft:goat_horn"))
    
def decode_use_cooldown(b64: str):
    data = bytearray(base64.b64decode(b64))
    ptr = 0

    cooldown = struct.unpack('>f', data[ptr:ptr+4])[0]
    ptr += 4

    has_group = data[ptr] == 1
    ptr += 1

    group = None
    if has_group:
        # Read VarInt
        shift = 0
        result = 0
        while True:
            byte = data[ptr]
            ptr += 1
            result |= (byte & 0x7F) << shift
            if not (byte & 0x80):
                break
            shift += 7
        
        group = data[ptr:ptr+result].decode('utf-8')
        ptr += result

    return cooldown, group
    
print(decode_use_cooldown(encode_use_cooldown(7.0, "minecraft:goat_horn")))
```
which produces
```
QOAAAAETbWluZWNyYWZ0OmdvYXRfaG9ybg==
(7.0, 'minecraft:goat_horn')
```

PR #1191, opened 2025-04-07
Add new RuntimeObsolete annotation used for checking if a field is obsolete at runtime via reflection

Axionize:


booky10 (approved):

booky10:
Idk if I'm 100% sure this is the correct name for this, but Idk a better one

PR #1192, opened 2025-04-07
make itemstack type getter always return type

ManInMyVan:
allows checking if the stack type is actually air without hacky stuff


booky10:
does this fix any relevant issues?

ManInMyVan:
> does this fix any relevant issues?

It fixes a bpu false with understacked items in grim

booky10 (approved):
change seems fine, i dont really see the point in doing this pointless empty-check before anyway

booky10:
`cachedIsEmpty` as a whole should probably revisited

PR #1204, opened 2025-04-18
Remove server version check for SelectBundleItem selectedItemIndex check

MachineBreaker:


MachineBreaker:
The version check is useless here, behavior doesnt change

booky10 (approved):
Thanks!

PR #1209, opened 2025-04-25
Fix several Fabric issues and add Fabric multi-version (1.14-1.21.5) support

Axionize:
Changes:
1. It supports Fabric 1.14-1.21.5 in the same jar (though online mode is broken in <= 1.20.4) 
2. It supports being run on the client, including on both the client itself and the internal server
3. It fixes issues with events not firing
4. It works with Carpet mod and doesn't error out on the NPC players by filtering the fake channels correctly
5. It fixes compatibility with ViaVersion and Krypton
6. Allows fetching custom server-side data by string
and more...

It is production proven and is used by Grim for the Grim Fabric port on many Fabric servers


umbrageodotus:
@retrooper ????
Gonna merge?

PR #1212, opened 2025-04-28
Add missing copy fields to spawn player wrapper

Beaness:


retrooper:
Thanks for your attentive proclivity to detail and thanks for your contribution.

PR #1250, opened 2025-06-18
Refactor TaskWrapper into interface and implementation classes

ShortyDev:
This commit message reflects the change already. It's a bit cleaner to handle TaskWrapper this way, it improves readbility, maintainability and extensibility for this part of the code.

Tests for this change are in my opinion not required, but I'm happy to provide logs or whatever you need if necessary.


retrooper:
Thank you for the PR!

retrooper:
We're sorry, but we won't accept this change for now, as it contains a breaking change, which could become an inconvenience for users depending on an external version of PacketEvents. However, we're willing to accept the proposed changes for 3.0.0.

PR #1251, opened 2025-06-19
Attempt fixing random login race condition injection issues

### LINKED ISSUE #1063: PacketEvents 2.0 failed to inject randomly
**Describe the bug**
Randomly players while joining will get kicked with "PacketEvents 2.0 failed to inject"

**Software brand**
Paper 1.20.4, paper 1.21 (this issue has been around a long time, since 2024-02-10 until present day)

**Plugins**
Protocollib, vulcan, grim

**How To Reproduce**
Randomly players will get disconnected when joining, its quite rare so hard to replicate, it has happened 191 times on my server out of ~150,000 player joins, here is a dump of all https://pastebin.com/GqBE0v0z

I assume this is because there is some kind of conflict/data race between protocollib, the shaded packet events in grim and the old shaded packet events in vulcan?

**Expected behavior**
Shouldn't kick them

**Additional context**
What causes this error? Is there some way to make packet events more fault tolerant rather than just kicking?, maybe some debug flag to println that I could enable and if it happens report back the debug data?

### LINKED ISSUE #1281: [HorriblePlayerLoginEventHack] You have plugins listening to the PlayerLoginEvent, this will cause re-configuration APIs to be unavailable: [LuckPerms, DeluxeMenus, GrimAC]
**Describe the bug**
[HorriblePlayerLoginEventHack] You have plugins listening to the PlayerLoginEvent, this will cause re-configuration APIs to be unavailable: [LuckPerms, DeluxeMenus, GrimAC]

**Software brand**
Leaf 1.21.7-8-b6972a7 (MC: 1.21.7)

**Plugins**
These plugins use PacketEvents

Image

booky10:
Basically rewrites how packetevents handles associating users with bukkit player objects

Should fix https://github.com/retrooper/packetevents/issues/1063
Also fixes more issues like https://github.com/retrooper/packetevents/issues/1036

Fixes https://github.com/retrooper/packetevents/issues/1281


booky10:
im pretty sure this is now broken on 1.21.7+ as of https://github.com/PaperMC/Paper/commit/7f609243907131c37d457aa367c04efbbd10f022, i'll try to look into this in a few days

booky10:
> im pretty sure this is now broken on 1.21.7+ as of [PaperMC/Paper@7f60924](https://github.com/PaperMC/Paper/commit/7f609243907131c37d457aa367c04efbbd10f022), i'll try to look into this in a few days

I've updated the PR to support Paper's configuration API changes; Paper will no longer create a ServerPlayer instance during the configuration phase, like vanilla does
This means plugins which use PacketEvents will also no longer have access to the Player object during the login/configuration protocol phases; though every packet sent during the play phase should have a Player object set

retrooper:
Looks good.

retrooper (approved):
Thanks!

PR #1265, opened 2025-06-26
fix spawn position for versions lower than 1.8 & protocol 4 support

Luuzzi:
If we send the packet in its current form, we get the error(on client side) I described below, and reading it will most likely result in a memory leak.

```
DecoderException: java.lang.IndexOutOfBoundsException: readerIndex(9) + length(4) exceeds writerIndex(11): UnpooledHeapByteBuf(rdx: 9, widx: 11, cap: 11)

```


Luuzzi:
Support for protocol 4 (1.7.2–1.7.5) could be useful for Velocity. I’ve tested it on my systems, and it generally works well.
If anyone will reports a bug, I’ll be glad to fix it.

retrooper:
Since it works, and we're not adding platform support for these versions, I don't see why not to accept this.

Luuzzi:
There was an incorrect bit shift — for example, StateTypes.GRASS_BLOCK was assigned the ID 32, which actually pointed to DEAD_BUSH.

1 - before
2 - after

Screenshot_0
Screenshot_1

api/src/main/java/com/github/retrooper/packetevents/protocol/player/ClientVersion.java
```diff
@@ -39,6 +39,7 @@
  * @since 1.6.9
  */
 public enum ClientVersion {
+    V_1_7_5(4),
```
booky10: Please rename it to `V_1_7_2`

booky10: Also, as you are modifying the first version we support, this has to be updated somewhere in this file

booky10: Additionally, many mapping files (`mappings` directory) reference 1.7.10 as it is the oldest supported version
You will need to update every reference in there

booky10 (changes requested):

api/src/main/java/com/github/retrooper/packetevents/protocol/player/ClientVersion.java
Luuzzi: > Please rename it to `V_1_7_2`

Oh, I see. I thought that they should be named according to the highest version. By analogy with 1.7.10. Thanks

Luuzzi: > Also, as you are modifying the first version we support, this has to be updated somewhere in this file

Found it, thanks!

Luuzzi: I think it's resolved

Luuzzi:
this is quite useful instead of hardcoding, will help to avoid multiple calls via switch -> case

jonesdevelopment (approved):
lgtm

Luuzzi:
@booky10, would it be possible to merge it?

overkidding:
Could you fix https://github.com/retrooper/packetevents/issues/1301 before this gets merged?

Luuzzi:
> Could you fix #1301 before this gets merged?

I think it’s done. Thanks!

Luuzzi:
<= 1.7.5

Screenshot_2

retrooper (approved):
Looks good.

PR #1267, opened 2025-06-26
Allow getting/setting json in ChatMessage

iiAhmedYT:
**Why?**

It's simple. I relocate adventure, It's way too painful to work with adventure & packetevents if its relocated.


retrooper:
Thanks.

Issue #1271, opened 2025-06-27
Add a way to get the current RemoteChatSession


ineanto:
**Is your feature request related to a problem? Please describe.**
When sending a new ```WrapperPlayServerPlayerInfoUpdate```, the Entry it requires has a ```RemoteChatSession``` field. There's currently no way of getting an instance of the player current chat session without intercepting a Packet first, which I would not like to have to do.

**Describe the solution you'd like**
A simple getter method to get via the ```User``` class, the ```RemoteChatSession``` of the player.

**Describe alternatives you've considered to solve your solution without
us adding this as a feature?**
As I said, I could intercept a packet containing the session but this is tedious.

**Additional context**
None

IronApollo:
+1

PR #1273, opened 2025-06-28
feat: add RemoteChatSession getter 

ineanto:
Partial fix to #1271 

This doesn't entirely fix the issue as PacketEvents requires a custom wrapper of ```RemoteChatSession``` around the ```WrapperPlayServerPlayerInfoUpdate.Entry``` field.


retrooper:
Thanks for the changes!

PR #1277, opened 2025-06-30
implement item components encode/decode in ItemStack

HSGamer:
This adds some code to support Item Component when encoding/decoding ItemStacks

Not all item components are supported

This is a preparation for us to gradually add support for all item components


HSGamer:
I think I've implemented enough component types for it to be usable. 
The main point of this PR is just for Item Body in the Dialog API. I think I've implemented enough to display most parts of the Item.

HSGamer:
Other developers can help by implementing the missing components.

HSGamer:
You can merge this now.

HSGamer:
To be honest, I'm too tired to implement all item components here. 
That's why I only implemented some necessary components for display item and some components that can be easily implemented using the available methods.
The current code heavily targets the items from packets, but not really those from NBT tags.

I'm gonna leave this PR open here and stop my work. Someone may continue the work, but at this point, I see no reason to continue this.

EDIT: As a result, all current TODOs are removed.

PR #1282, opened 2025-07-04
Forced order personal packet transformers

semenishchev:
Adds a PacketTransformer class which accepts PacketSendEvent and PacketReceiveEvent before and after any listeners are invoked. 

Primary reason is to allow to easily force order of some packet events, for example allow proper coordinate obfuscating/deobfuscating before packets hit the anticheat.

Secondary reason is probably to save a few nanoseconds in processing times for plugins which need to handle all packets (said coordinate obfuscators) to avoid doing something like PlayerData.get(event.getUser().getUUID()) rather directly giving ability to set transformer and pass user data into it.

Initially I wanted to make a map of transformers similar to netty pipeline, but didn't want to overcomplicate things, + there's always going to be a bold plugin which thinks that it's the most important one. 

That's how I use it in my coordinate obfuscator
```java
private class UserListener extends PacketListenerAbstract {
        @Override
        public void onUserLogin(UserLoginEvent event) {
            User user = event.getUser();
            user.setTransformer(new ObfuscatorTransformer(user, PlayerData.get(user.getUUID())));
        }
  }
  
  public class Transformer extends PacketTransformer {
        private final PlayerData data;

        public Transformer(User user, PlayerData data) {
            super(user);
            this.data = data;
        }

        @Override
        public void onPostSend(PacketSendEvent event) {
            data.obfuscate(event);
        }

        @Override
        public void onPreReceive(PacketReceiveEvent event) {
           data.deobfuscate(event);
        }
    }
  ```

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/protocol/PacketTransformer.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/protocol/PacketTransformer.java
@@ -1,0 +1,18 @@
+package com.github.retrooper.packetevents.protocol;
+
+import com.github.retrooper.packetevents.event.PacketReceiveEvent;
+import com.github.retrooper.packetevents.event.PacketSendEvent;
+import com.github.retrooper.packetevents.protocol.player.User;
+
+public class PacketTransformer {
+   protected final User user;
+
+   public PacketTransformer(User user) {
+      this.user = user;
+   }
+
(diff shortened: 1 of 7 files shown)


retrooper:
Please make a GitHub issue on the repository to propose a change or feature before working on a PR. Using the packet listener priorities should be sufficient. Thanks for the PR, though.

PR #1286, opened 2025-07-08
Revert "Refactor TaskWrapper into interface and implementation classes"

retrooper:
Reverts retrooper/packetevents#1250 due to breaking change


ShortyDev:
images

PR #1306, opened 2025-07-22
ViaNBT, SpongeGradle, MockBukkit and shadow update

rafi67000:
I updated ViaNBT (they relocated the path in 5.0), SpongeGradle, MockBukkit (also relocation, and renamed few classes)
it's probably one of last updates in shadow 8.x release cycle and it should be updated to 9.0 once it's released as stated [here](https://github.com/GradleUp/shadow/releases/tag/9.0.0-rc1)
Tests run successfully and packetevents compiled successfully


retrooper:
Thanks for the insight and the PR!

PR #1308, opened 2025-07-29
Fix NBTCompound.java getListOrThrow always returning null

Moulberry:
Fix NBTCompound.java getListOrThrow always returning null


retrooper:
Thanks for the PR!

PR #1309, opened 2025-07-29
Fix NPE in SpigotChannelInjector#isPlayerSet

Moulberry:
Prevent SpigotChannelInjector#isPlayerSet from throwing NPE when encoder/decoder isn't present


retrooper:
Thanks for the changes.

retrooper:
Thanks for the changes.

PR #1323, opened 2025-08-06
Allow PacketEvents to override sign text limit per player

GigaZelensky:
This PR lets PacketEvents override the default sign text limit per player. Minecraft normally caps sign lines at 384 characters, but right now that limit is hardcoded. With this change, other plugins can set custom limits for each player as needed.

It's a straightforward change and doesn't affect anything unless explicitly used by another plugin. The implementation is thread-safe and efficient.

This is not really necessary to me anymore, if it doesn't fit in with PacketEvents it can be removed.


AoElite:
@GigaZelensky I don't think this is a good way to do this. Maybe just make another constructor that allows you to pass in the max limit there instead.

GigaZelensky:
> @GigaZelensky I don't think this is a good way to do this. Maybe just make another constructor that allows you to pass in the max limit there instead.

Yeah I ended up doing something else that doesn't require this so this is not really necessary anymore. Plugins can do their own version of this anyway

PR #1330, opened 2025-08-13
Fix components for 1.9-1.11.2 versions

shChallenger:
This PR fixes an issue where components are encoded in a format not supported by these versions.

In these versions, the client disconnects when receiving a plain text component (i.e., a JSON containing only a string without a `text` key, style, or color), resulting in the following error:

image

**Background:**
Kyori Components generate a simplified JSON for plain text components. For example, instead of giving `{"text": "§csome text" }`, they return just `"§csome text"`, as the Minecraft client is expected to support this format.

Adding a default color to the components forces them to be encoded in the standard format, which is the approach used here.

Starting with version 1.9, Minecraft switched from using `Gson#fromJson()` to `JsonReader` to better handle lenient JSON strings. However, at that time, a bug in the Gson library caused Lenient exceptions when parsing plain JSON strings as shown above.

This issue was fixed in version 1.12 by upgrading to Gson 2.8.0, which resolves the problem.

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/util/adventure/AdventureSerializer.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/util/adventure/AdventureSerializer.java
@@ -230,6 +230,7 @@ public final class AdventureSerializer implements NbtEncoder<Component>, NbtDeco
                                 .values(JSONOptions.byDataVersion().at(0))
                                 .value(JSONOptions.EMIT_HOVER_EVENT_TYPE, JSONOptions.HoverEventValueMode.BOTH)
                                 .value(JSONOptions.SHOW_ITEM_HOVER_DATA_MODE, JSONOptions.ShowItemHoverDataMode.EMIT_EITHER)
+                                // Ensure compatibility with versions 1.9 to 1.11.2
                                 .value(JSONOptions.EMIT_COMPACT_TEXT_COMPONENT, false);
                         if (this.version.isNewerThanOrEquals(ClientVersion.V_1_16)
                                 && !PacketEvents.getAPI().getSettings().shouldDownsampleColors()) {
--- a/api/src/main/java/com/github/retrooper/packetevents/util/adventure/AdventureSerializer.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/util/adventure/AdventureSerializer.java
@@ -229,7 +229,8 @@ public final class AdventureSerializer implements NbtEncoder<Component>, NbtDeco
                         builder
                                 .values(JSONOptions.byDataVersion().at(0))
                                 .value(JSONOptions.EMIT_HOVER_EVENT_TYPE, JSONOptions.HoverEventValueMode.BOTH)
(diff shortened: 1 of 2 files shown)


shChallenger:
**Note:**
This issue mainly affects the following packets: `WrapperPlayServerOpenWindow`, `WrapperPlayServerChatMessage`, and `WrapperPlayServerDisconnect`.

With the standard Spigot API, GUI titles are encoded as a single string using `§` for colors. As a result, intercepting `WrapperPlayServerOpenWindow` will cause clients on these versions to be kicked when opening any GUI.

Similarly, disconnecting a user with an unstyled message will trigger the same error.

booky10:
You can just set the `JSONOptions.EMIT_COMPACT_TEXT_COMPONENT` option [here](https://github.com/retrooper/packetevents/blob/24bf010d9817545de6e3e8c62132e6770d4d1fa5/api/src/main/java/com/github/retrooper/packetevents/util/adventure/AdventureSerializer.java#L229-L246) instead of modifying the component's color

shChallenger:
> You can just set the `JSONOptions.EMIT_COMPACT_TEXT_COMPONENT` option [here](https://github.com/retrooper/packetevents/blob/24bf010d9817545de6e3e8c62132e6770d4d1fa5/api/src/main/java/com/github/retrooper/packetevents/util/adventure/AdventureSerializer.java#L229-L246) instead of modifying the component's color

**EDIT:** Removing the compact component is probably the better approach, but it seems we cannot check the client version in `AdventureSerializer`. As a result, other versions would no longer benefit from reduced components.

booky10:
> > You can just set the `JSONOptions.EMIT_COMPACT_TEXT_COMPONENT` option [here](https://github.com/retrooper/packetevents/blob/24bf010d9817545de6e3e8c62132e6770d4d1fa5/api/src/main/java/com/github/retrooper/packetevents/util/adventure/AdventureSerializer.java#L229-L246) instead of modifying the component's color
> 
> **EDIT:** Removing the compact component is probably the better approach, but it seems we cannot check the client version in `AdventureSerializer`. As a result, other versions would no longer benefit from reduced components.

There is a `version` field, which basically is the client version

shChallenger:
> > > You can just set the `JSONOptions.EMIT_COMPACT_TEXT_COMPONENT` option [here](https://github.com/retrooper/packetevents/blob/24bf010d9817545de6e3e8c62132e6770d4d1fa5/api/src/main/java/com/github/retrooper/packetevents/util/adventure/AdventureSerializer.java#L229-L246) instead of modifying the component's color
> > 
> > 
> > **EDIT:** Removing the compact component is probably the better approach, but it seems we cannot check the client version in `AdventureSerializer`. As a result, other versions would no longer benefit from reduced components.
> 
> There is a `version` field, which basically is the client version

I thought it used the client version, but after some tests I saw it uses the server version when getting the `AdventureSerializer` from the `PacketWrapper`:

https://github.com/retrooper/packetevents/blob/24bf010d9817545de6e3e8c62132e6770d4d1fa5/api/src/main/java/com/github/retrooper/packetevents/util/adventure/AdventureSerializer.java#L55-L57

PR #1333, opened 2025-08-18
Always try to fetch netty version

Beaness:
If netty-common or netty-all is not present it will fallback by fetching any possible netty lib version


rafi67000:
I don't think netty will work without netty-common

retrooper:
Have you experienced any issues, and has this code _actually_ solved something for you?

Beaness:
Yes I was using a spigot fork which had only netty-handler in the versions file (even though multiple netty modules were present), which caused packetevents to not fetch any netty version.
This happens if the netty module versions aren't properly merged into the io.netty.versions.properties file

(Another example: velocity only has netty-codec)
image

PR #1341, opened 2025-09-05
Refactor properties and equipment processing to dedicated method

AoElite:
Refactor equipment reading into a dedicated method, invoked only when needed. This allows subclasses to override the implementation or skip reading altogether, improving customization and optionally enabling better performance in specific use cases.


AoElite:
Added a method for reading properties from the update attribute packet

retrooper (approved):
I approve of these changes. They're reasonable.

PR #1358, opened 2025-09-29
Fix CLICK_WINDOW wrapper write length

Jyguy:
The previous `PacketWrapper#writeOptionalMap()` incorrectly did not write a boolean to indicate the presence of the `HashedStack`. This PR fixes this by using the normal `PacketWrapper#writeMap()` and using `HashedStack#writeOptional()` instead, which correctly writes false if the optional is empty.


booky10:
Thanks for the fix!

PR #1362, opened 2025-10-01
Gradle update

rafi67000:
Updated gradle, shadow, mod-publish, loom, mockbukkit
Replaced Tasks#create with Tasks#registering (that's the preferred way in gradle 9.x)
~~I also reverted the commit that caused github actions to not publish shadowed jars~~
~~I changed paths in the workflow so shadow jars in /build/libs will also be picked up by upload action.~~
I added a step in the workflow to copy jars from build/libs (Fixes regression after a861066)
MockBukkit doesn't ship paper-api anymore

I'll open another PR when I manage to make shadow 9 work


rafi67000:
also I was wondering can the fastutil be updated? in buildSrc

rafi67000:
yeah now they are upload the way they should be

booky10:
> also I was wondering can the fastutil be updated? in buildSrc

As long as the build process still works, everything in `buildSrc` can be updated - everything in there is just for building/testing/publishing packetevents

PR #1368, opened 2025-10-07
fix: resolve NoSuchMethodError in login event with isAllowed() check

### LINKED ISSUE #1367: Could not pass event PlayerConnectionValidateLoginEvent
**Describe the bug**
When a player joins the server, a `NoSuchMethodError` occurs due to a call to the *non-existent* `getKickMessage()` method in Paper. This issue only if PacketEvents and Kyori Adventure are relocated

The problem in version 2.10.0-SNAPSHOT https://github.com/retrooper/packetevents/blob/2.0/spigot/src/main/java/io/github/retrooper/packetevents/bukkit/InternalPaperJoinListener.java#L71

I am aware that PacketEvents does not *full* support relocation/shading and discourages the practice. However, I have specific reasons for this. It is important to note that this issue was not present in previous versions.

**Software brand**
Paper 1.21.9

**Plugins**
My custom plugin (FlectonePulse)

**How To Reproduce**
1. Relocate PacketEvents and Kyori Adventure
2. Join the server

**Expected behavior**
No errors

**Additional context**
Logs https://mclo.gs/BueLM9Q

TheFaser:
Fix #1367

### DIFF
--- a/spigot/src/main/java/io/github/retrooper/packetevents/bukkit/InternalPaperJoinListener.java
+++ b/spigot/src/main/java/io/github/retrooper/packetevents/bukkit/InternalPaperJoinListener.java
@@ -68,7 +68,7 @@ public class InternalPaperJoinListener implements Listener {
 
     @EventHandler(priority = EventPriority.MONITOR)
     public void onLogin(PlayerConnectionValidateLoginEvent event) {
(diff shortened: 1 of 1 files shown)


booky10 (approved):
Thanks for fixing this, I somehow missed this method 👍

PR #1370, opened 2025-10-07
update adventure version and add support object components

TheFaser:


TheFaser:
about `BackwardCompatUtil.IS_4_25_0_OR_NEWER`, I'm not sure if it's necessary. I think we can only check the client version, but I left it

Bakinu:
good good accept this pull request pls

booky10 (commented):
I refactored the profile parsing into our more generic `ItemProfile` class (`ResolvableProfile` in vanilla) and added some abstract codec implementations which vanilla uses everywhere

Additionally, I also updated our "patched" adventure gson serializer to properly support the new object component (this may also fix some shadow color related issues?)
I dont like how packetevents currently interferes with adventure, but until we can do a breaking change, this will likely remain the case

booky10:
Thanks for your contribution!

PR #1374, opened 2025-10-19
Fix 1.21.9/1.21.10 particle types

### LINKED ISSUE #1373: Ender Dragon Breath Causing Crash + Containers Bug
**Describe the bug**
Ender Dragon Breath causing crash - When fighting the Enderdragon, when he shoots his dragons breath it causes cliented to be crashed immediately when trying to render the particles - Stacktrace added below

Containers Bug - Players will become unable to open / interact with blocks like doors and containers unless they either relog or shift. They seem to get stuck in a state of being shifted while not shifted.

**Software brand**
Purpur 1.21.10 Build 2511

**Plugins**
Screenshot below. Packetevents is missing because it was removed to fix the bug

**How To Reproduce**
Ender Dragon bug - Spawn ender dragon, wait til it uses breath

Container Bug - Not sure how to specifically recreate it.. Will update with details if found.

**Expected behavior**
No crash on ender dragon breath

Able to open containers / interact with blocks normally

Tested using Packetevents released on 10/17/2025

Image

[disconnect-2025-10-16_21.33.20-client.txt](https://github.com/user-attachments/files/22986361/disconnect-2025-10-16_21.33.20-client.txt)

booky10:
Fixes #1373

Draft, because for some particle types this is a horrible solution and breaks backwards compat in some cases


retrooper:
Thanks for the quick correction. I couldn't find a better fix to this, but I don't think many developers will face significant issues with this change.

PR #1376, opened 2025-10-24
.

ghost:


jonesdevelopment (approved):

booky10:
Minecraft itself doesn't see them as solid, is there are any reason for marking them as solid?
In the case of collisions, hanging signs are also not marked as solid and still have collisions

AnnotationVisitor:
> Minecraft itself doesn't see them as solid, is there are any reason for marking them as solid? In the case of collisions, hanging signs are also not marked as solid and still have collisions

Solid because that's what makes sense, it has almost the same collision box as happy ghast, which is marked solid, not to mention solid=true is what works with grim

booky10:
No, it doesn't make sense, these flags are copied from vanilla

Why is this required for grim? The solid state isn't really used client-side as far as I know
I hope Grim isn't using this to determine whether a block has collisions or not, because thats a separate flag in vanilla (which is not copied by packetevents and is also different for e.g. hanging signs)

AnnotationVisitor:
> No, it doesn't make sense, these flags are copied from vanilla
> 
> Why is this required for grim? The solid state isn't really used client-side as far as I know I hope Grim isn't using this to determine whether a block has collisions or not, because thats a separate flag in vanilla (which is not copied by packetevents and is also different for e.g. hanging signs)

https://github.com/GrimAnticheat/Grim/blob/1b47e31776cd4c4386fd7667f72701d00f4a1156/common/src/main/java/ac/grim/grimac/utils/collisions/CollisionData.java#L1149

AnnotationVisitor:
@ManInMyVan thoughts? tbh the issolid optimization in getData is kind of hacky, i'd just use a reference2objectmap with state type key and collision value

ManInMyVan:
> @ManInMyVan thoughts?

add an override

booky10:
You are free to open a PR adding this "noCollision" flag to packetevents
If you do so, please generate the data and don't do it manually

PR #1380, opened 2025-10-26
Add constructor with data array to ListPalette

Rothes:
No longer needs [clumpy way](https://github.com/Rothes/ESU/blob/17c8960d06a7fac2858a0c85ec7fce15a66e2f2e/bukkit/version/v1_18/src/main/kotlin/io/github/rothes/esu/bukkit/module/networkthrottle/chunkdatathrottle/v1_18/ChunkDataThrottleHandlerImpl.kt#L866) to do it

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/protocol/world/chunk/palette/ListPalette.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/protocol/world/chunk/palette/ListPalette.java
@@ -27,6 +27,8 @@ package com.github.retrooper.packetevents.protocol.world.chunk.palette;
 import com.github.retrooper.packetevents.protocol.stream.NetStreamInput;
 import com.github.retrooper.packetevents.wrapper.PacketWrapper;
 
+import java.util.Arrays;
+
 /**
  * A palette backed by a List.
  */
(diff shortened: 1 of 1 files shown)


api/src/main/java/com/github/retrooper/packetevents/protocol/world/chunk/palette/ListPalette.java
```diff
@@ -63,6 +64,12 @@ public ListPalette(int bitsPerEntry, PacketWrapper<?> wrapper) {
         this.nextId = paletteLength;
     }
 
+    public ListPalette(int bitsPerEntry, int[] data) {
+        this.bits = bitsPerEntry;
+        this.data = data;
```
booky10: please verify the length of the data matches the bits-per-entry parameter

booky10 (changes requested):

api/src/main/java/com/github/retrooper/packetevents/protocol/world/chunk/palette/ListPalette.java
booky10: I'm pretty sure the data array length has to exactly match   the bits, though the array doesn't need to be fully filled

Rothes: I would do a `Arrays.copyOf(data, expectedSize)` if the data array length is lower than the bits, does it make sense or I should throw an exception?

booky10: Copying the array is fine :+1:
But if you do that, please always copy/clone it, so the constructor behaves consistently

Rothes: It's the behavior now.

booky10 (approved):

booky10:
Thanks for your PR!

PR #1382, opened 2025-10-27
Fix nbt list unwrapping logic

### LINKED ISSUE #1377: [BUG] [FABRIC 1.21.9+] Incorrect reading of player_head
**Describe the bug**
PacketEvents doesn't work correctly on Fabric 1.21.9+. When retrieving a message from SYSTEM_CHAT_MESSAGE, the player_head will be incorrect and won't contain profile properties. This problem doesn't occur on Bukkit servers. If you not reading messages from SYSTEM_CHAT_MESSAGE, the issue won't happen.

**Software brand**
Fabric 1.21.9+

**Plugins**
PacketEvents 2.10.0 and custom mod

**How To Reproduce**

Create fabric mod with
```java
@Override
public void onPacketSend(PacketSendEvent event) {
    if (event.getPacketType() == PacketType.Play.Server.SYSTEM_CHAT_MESSAGE) {
        new WrapperPlayServerSystemChatMessage(event);
    }
}
```

1. Start the fabric server with online_mode=true and your mod
2. Join with the 1.21.9+ vanilla minecraft client
3. /fetchprofile name <name>
4. See incorrect player_head

Image

**Expected behavior**
The issue is clearly related to reading components on Fabric. The player_head should display correctly

TheFaser:
Fix #1377

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/protocol/component/builtin/item/ItemProfile.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/protocol/component/builtin/item/ItemProfile.java
@@ -21,7 +21,6 @@ package com.github.retrooper.packetevents.protocol.component.builtin.item;
 import com.github.retrooper.packetevents.manager.server.ServerVersion;
 import com.github.retrooper.packetevents.protocol.nbt.NBT;
 import com.github.retrooper.packetevents.protocol.nbt.NBTCompound;
-import com.github.retrooper.packetevents.protocol.nbt.NBTList;
 import com.github.retrooper.packetevents.protocol.nbt.NBTString;
 import com.github.retrooper.packetevents.protocol.player.PlayerModelType;
 import com.github.retrooper.packetevents.protocol.util.NbtCodec;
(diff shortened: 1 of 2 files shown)


booky10:
Did you test this actually changes something? If working correctly, `return this.propertyList.decode(nbt, wrapper);` (see immediately below your changes) should also handle nbt lists

TheFaser:
> Did you test this actually changes something? If working correctly, `return this.propertyList.decode(nbt, wrapper);` (see immediately below your changes) should also handle nbt lists

Yes, I've tested this on Fabric. I also thought that the lists should be processed already, but it doesn't happen on Fabric

TheFaser:
I tested it on Fabric 1.21.10 and on Paper 1.21.9, it works without errors

booky10:
@TheFaser Could you please test again with my additional changes? I fixed tag unwrapping no longer working after your changes

TheFaser:
> @TheFaser Could you please test again with my additional changes?

yes, it works

booky10:
Thanks!

PR #1389, opened 2025-11-01
Validate Sponge Adventure version & fix logging

SamB440:
I keep having issues where the adventure version between Sponge and PacketEvents is mismatched since we aren't meant to shade on Sponge.

This adds automatic detection of the Adventure version Sponge and PacketEvents is using and outputs a useful log message if they aren't the same.
image

I also fixed logging by using the actual plugin's logger, not trying to use systemSubject (colours using that just don't work). It now basically matches what the Fabric one does.

### DIFF
--- a/sponge/build.gradle.kts
+++ b/sponge/build.gradle.kts
@@ -13,8 +13,10 @@ repositories {
     maven("https://repo.viaversion.com/")
 }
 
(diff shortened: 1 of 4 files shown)


retrooper:
Interesting change. Thanks for the PR.

PR #1407, opened 2025-11-30
fix: thread save packet type init

NgLoader:
This pull request addresses a concurrency issue in [PacketType::prepare](https://github.com/retrooper/packetevents/blob/2.0/api/src/main/java/com/github/retrooper/packetevents/protocol/packettype/PacketType.java#L440) that can result in a disconnect for players. Under heavy parallel load, prepare may be invoked concurrently, but the method is not thread safe. This can lead to to the following exception:

```
java.lang.ArrayIndexOutOfBoundsException: Index 24 out of bounds for length 23
    at com.github.retrooper.packetevents.protocol.packettype.PacketType$Play$Client.loadPacketIds(PacketType.java:1202)
    at com.github.retrooper.packetevents.protocol.packettype.PacketType$Play$Client.load(PacketType.java:1224)
    at com.github.retrooper.packetevents.protocol.packettype.PacketType.prepare(PacketType.java:441)
    ...
```

This failure was observed when integrating PacketEvents with GrimAC, where multiple players joining at the exact same moment triggered concurrent calls to prepare before packet types were fully initialized.

To resolve this, the PR introduces a simple exclusive lock around the prepare method. The lock ensures that any concurrent calls wait until packet IDs are fully loaded, preventing data races and guaranteeing consistent initialization.

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/protocol/packettype/PacketType.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/protocol/packettype/PacketType.java
@@ -364,7 +364,8 @@ import java.util.Map;
 
 public final class PacketType {
 
-    private static boolean PREPARED = false;
+    private static volatile boolean PREPARED = false;
+    private static final Object PREPARE_LOCK = new Object();
 
     //TODO UPDATE Update packet type mappings (clientbound pt. 1)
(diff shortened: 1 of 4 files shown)


booky10:
Thanks for the fix!

PR #1409, opened 2025-12-07
fix: correct default buffer size to match comment

groundbreakingmc:
The code previously used 1 << 20 bytes (1MiB) for DEFAULT_MAX_SIZE, but the comment stated 2MiB. This updates the value to 2 << 20 bytes to match the comment and intended behavior.


groundbreakingmc:
This also fixes one new crash

booky10 (approved):

PR #1411, opened 2025-12-08
ResourceLocation refactor

AoElite:
- Refactored the `ResourceLocation` class to use the `Key` type from the Adventure library
- Fixed StateTypes having invalid key patterns in their ResourceLocation


retrooper (approved):
The author reduces memory allocations and code complexity. I would approve of these changes.

PR #1413, opened 2025-12-08
No injection condition called twice

Emibergo02:


jonesdevelopment:
I'm confused. One of the checks is for the _decoder_, and one is for the _encoder_. There shouldn't be any duplicates here.

PR #1414, opened 2025-12-08
Implement a custom Int2Int map for MapPalette

### LINKED ISSUE #1412: Memory consideration
**Is your feature request related to a problem? Please describe.**
MapPalette uses a java.util.HashMap with an Integer value.
GrimAC likes keeping millions of instances loaded.

**Describe the solution you'd like**
As the comment above the field states,
```java
// TODO: Can we use fastutils here?
private final HashMap<Object, Integer> stateToId = new HashMap<>();
```
FastUtil would work fine here, basically any implementation of an Object2IntMap.

**Describe alternatives you've considered to solve your solution without
us adding this as a feature?**
It's a simple change I would've submitted a pull request for if it weren't for the 5 different platforms I'd have to test and worry about shading.

Vrganj:
Replaces a very unnecessary HashMap<Object, Integer> with a much lighter custom Int2Int map.
Works fine from my tests, might warrant a unit test.

Closes #1412

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/protocol/world/chunk/palette/Int2IntHashMap.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/protocol/world/chunk/palette/Int2IntHashMap.java
@@ -1,6 +1,7 @@
 package com.github.retrooper.packetevents.protocol.world.chunk.palette;
 
 import org.jetbrains.annotations.ApiStatus;
+import org.jspecify.annotations.NullMarked;
 
 import java.util.Arrays;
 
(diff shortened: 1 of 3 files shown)


Vrganj:
Did a rough comparison with GrimAC which stores all the chunks sent to players. The heap dump I took before the patch shows 33% of the total chunk memory data was used by palettes. After the patch, the usage dropped to about 17%, resulting in about a 20% memory improvement per average chunk in my case.

I'm aware they use a fork of packetevents so I applied the patch to their fork to test. Considering they seem to keep upstream commits up to date, I thought it'd be better to just merge the change here.

booky10 (changes requested):
Please refactor this new Map into a separate class (and mark it as `@ApiStatus.Internal` then)

Also, you mentioned copying existing libraries in your issue - was this copied from somewhere or did you write this on your own?

Vrganj:
Mostly wrote it myself, but took great inspiration from https://github.com/aeron-io/agrona/blob/master/agrona/src/main/java/org/agrona/collections/Int2IntHashMap.java which should be compliant licensing-wise. The map should probably be renamed to something like UInt2UIntMap considering I chose to use -1 as the empty key value.

booky10 (approved):

booky10:
Thanks for the PR and sorry that it took so long to merge!

PR #1417, opened 2025-12-11
Add comment to indicate the mojang name of the packet the wrapper is handling

vadim-soude:
This is some parity comment in regard of ClientboundLevelEventPacket


booky10:
Thanks!

PR #1431, opened 2026-01-05
Add new offset(Blockface face, int i) method to Vector3f/d/i classes

Axionize:
Method like offset(BlockFace face) but for integer multiples.


booky10 (approved):

retrooper (changes requested):
It could help readers if the offset function were documented, particularly explaining the purpose of the integer 'i' argument.

retrooper (approved):
I approve of these changes now. Thanks for the revision.

PR #1432, opened 2026-01-05
BlockTags for dead and all corals

Axionize:


api/src/main/java/com/github/retrooper/packetevents/protocol/world/states/defaulttags/BlockTags.java
```diff
@@ -387,14 +387,22 @@ public class BlockTags {
      * Unofficial tag for all glass panes
      */
     public static final BlockTags GLASS_PANES = bind("glass_panes");
+    /**
+     * Unofficial tag for all dead coral plants
+     */
+    public static final BlockTags DEAD_CORAL_PLANTS = bind("dead_coral_plants");
+    /**
+     * Unofficial tag for all dead coral plants
```
booky10, line 395:
```suggestion
     * Unofficial tag for all dead corals
```

booky10:
@Axionize why did you close this?

Axionize:
Because its wrong and I don't want to shove more and more unofficial tags this way that don't align with their vanilla counerparts. If you want I can reopen and think about this again but given vanilla's corals tags also includes fans but not wall fans and that there's no coral_fans and and coral_plants and that language makes it unclear if it includes wall fans

yeah it just got messy fast

PR #1456, opened 2026-02-07
Handle compression only if the server sends the compression packet

### LINKED ISSUE #1455: Performance degradation on servers with compression disabled
**Describe the bug**
When you are using PacketEvents on servers below version 1.19.4, or even on newer versions that do not use Paper as a base, and the server has compression disabled, PacketEvents keeps searching for the compressor in the pipeline for every packet sent to the player. This causes GC pressure (since Netty internally creates a new list every time you call `ChannelPipeline#names`), and high CPU usage on servers with a high packet rate.

**Software brand**
git-PaperSpigot-445 (MC: 1.8.8)

**Plugins**
packetevents

**How To Reproduce**
1 - Create a 1.8.8 server.
2 - Add a debug statement on this line: [https://github.com/retrooper/packetevents/blob/1b0a8bf8de9030fbc0a8ab97edab6e20d852f188/spigot/src/main/java/io/github/retrooper/packetevents/injector/handlers/PacketEventsEncoder.java#L241](https://github.com/retrooper/packetevents/blob/1b0a8bf8de9030fbc0a8ab97edab6e20d852f188/spigot/src/main/java/io/github/retrooper/packetevents/injector/handlers/PacketEventsEncoder.java#L241)
3 - Put PacketEvents with the debug into the `plugins` folder.
4 - Change the `network-compression-threshold` option to `-1`.
5 - Join the server and observe that the debug is constantly triggered because `compress` will never be found by PacketEvents.

**Expected behavior**
Only handle compression if, and after, the server sends the SET_COMPRESSION packet to the player.

**Screenshots**

Image

**Additional context**
I haven’t tested it, but probably Sponge suffers from the same issue, since it’s the same code.

ItsClairton:
fixes #1455


retrooper:
I'm busy with finals until end April, so it may take some time until this and other PRs are reviewed by *me*. Do you think you could resolve the merge conflicts in the meantime?

booky10:
Thanks for your PR and sorry for the long wait!

PR #1460, opened 2026-02-17
Instantly run task after write operation

Beaness:
Right now it's possible other packets get written before the promise is completed, If you want to write a packet right after another packet using getTasksAfterSend it's possible another packet gets written inbetween. This solves it for spigot (sponge needs a different type of implementation, probably overwriting the whole write function to execute the tasks).


booky10 (approved):
Thanks for the fix, I also applied your changes to sponge and fabric
This promise-based handling is because our spigot/paper outbound handler used to be a `MessageToMessageEncoder` and never was removed when we stopped using that

Sorry for the long wait!

retrooper:
Thank you for the PR.

PR #1468, opened 2026-02-25
Add system property to allow longer chat messages on legacy versions

Beaness:
Similar to
https://github.com/PaperMC/Velocity/blob/6aff78728c267bbb67b07298b8cf4580fb79fc6f/proxy/src/main/java/com/velocitypowered/proxy/protocol/packet/chat/legacy/LegacyChatPacket.java#L49

LunarClient Apollo API allows you to override the max chat length message for older mc versions. Without this change plugins that listen for chat packets break.

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/wrapper/play/client/WrapperPlayClientChatMessage.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/wrapper/play/client/WrapperPlayClientChatMessage.java
@@ -35,6 +35,18 @@ import java.util.Optional;
  * This packet is used to send a chat message to the server.
  */
 public class WrapperPlayClientChatMessage extends PacketWrapper<WrapperPlayClientChatMessage> {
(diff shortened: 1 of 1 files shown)


NoJokeFNA:
Why don't we keep the current design?
https://github.com/retrooper/packetevents/blob/2.0/api/src/main/java/com/github/retrooper/packetevents/protocol/nbt/NBTLimiter.java#L28

And I'm not sure if this will even get merged just because the "LunarClient Apollo API" allows you to change some values.

Beaness:
> Why don't we keep the current design? https://github.com/retrooper/packetevents/blob/2.0/api/src/main/java/com/github/retrooper/packetevents/protocol/nbt/NBTLimiter.java#L28

Fixed
> And I'm not sure if this will even get merged just because the "LunarClient Apollo API" allows you to change some values.

Seems like a minor addition that isn't super hard to maintain, the default value also doesn't change with this pr.

booky10 (approved):

PR #1483, opened 2026-04-06
Fix asBukkitCopy reflection for MC 26.1.1+

roggy666:
MC 26.1.1 introduced a new CraftItemStack.asBukkitCopy(ItemStackTemplate) overload. The index-based method lookup (getMethod with index 0) could resolve to the wrong overload depending on JVM method ordering, causing "argument type mismatch" errors for item-holding disguises.

Use getMethodExact with explicit NMS_ITEM_STACK_CLASS parameter type to always resolve the correct asBukkitCopy(ItemStack) method, with a fallback to the old index-based lookup for older servers.


retrooper (approved):
I've reviewed and tested your changes. Thanks for your contribution to PacketEvents.

PR #1486, opened 2026-04-10
add copy constructor for WrapperPlayServerPlayerInfoUpdate$PlayerInfo

ManInMyVan:
meant for plugins that need to modify this without modifying the original and don't want to update the constructor every time it changes


booky10 (approved):
looks good

PR #1488, opened 2026-04-12
Fix PaintingVariant read/write for >=1.21

ytnoos:


booky10 (approved):
I also added hte missing title/author fields

booky10:
Thanks for the PR!

PR #1496, opened 2026-04-16
Remove username length limit

Rothes:
Remove this limit for offline-mode servers, which actually accept 16+ length usernames.


jonesdevelopment:
This is a standard vanilla limit present in both proxies (e.g., Velocity) and regular Minecraft server software.

I don't think this limit should be removed. If you modify your software to this extent, you may as well just fork PE.

PR #1504, opened 2026-04-25
flexible listener priorities

ieatglu3:
This rewrites the listener priority system in a way that is more robust, including supporting custom integer priorities.

For maximum backwards compatibility, this does not remove the legacy `PacketListenerPriority`, nor its usages, instead just deprecates it without sacrificing existing functionality.

The modern listener priority class is now `ListenerPriority`, with the same default/predefined priority values as its legacy counterpart. 

User's can continue to use existing priorities, or create their own using the new system.

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/event/PacketListenerCommon.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/event/PacketListenerCommon.java
@@ -18,10 +18,6 @@
 
 package com.github.retrooper.packetevents.event;
 
-import java.lang.reflect.Method;
-import java.util.List;
-import java.util.Map;
-
 /**
  * Abstract packet listener.
  *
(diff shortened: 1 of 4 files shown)


rafi67000:
Why not reuse the PacketListenerPriority?

ieatglu3:
> Why not reuse the PacketListenerPriority?

This is a breaking change; the class type, and structure would have changed. This enum has remained the same since 1.0 (5 years). Unless breaking plugins... Isn't a concern? Then sure, It would make more sense to reuse it.

booky10:
@ieatglu3 can you activate this? want to push a few changes + formatting fixes
image

ieatglu3:
> @ieatglu3 can you activate this? want to push a few changes + formatting fixes
> 
> image
> 
> 

there

booky10:
Thanks

api/src/main/java/com/github/retrooper/packetevents/event/PacketListenerCommon.java
```diff
@@ -18,29 +18,53 @@
...
+        return this.priority.legacyType();
+    }
+
+    /**
+     * @return the priority of this packet listener.
+     */
+    public ListenerPriority priority() {
+        // no point in providing a backwards compat if method returns the modern type
+        // + the new type has to be exposed in one way or another (the event manager needs it), cant get around this one :(
+        return this.priority;
```
booky10, line 67:
```suggestion
        // backwards compat, developers may have overrides
        return ListenerPriority.fromLegacy(this.getPriority());
```
this needs to check for overrides of `#getPriority()`

booky10 (changes requested):

api/src/main/java/com/github/retrooper/packetevents/event/PacketListenerCommon.java
ieatglu3: bro?? This always just returns a predefined value; custom priorities become useless. Even then, this Is an overcompensation, If someone overrode `#getPriority`, then this just means their **legacy** override would be ignored in favour for what was passed into the constructor. The `priority` has always been `final` meaning if someone overrode it's **!!! getter !!!** `#getPriority`, the override inherently becomes unsafe.

PR #1505, opened 2026-04-28
Rework adventure handling

### LINKED ISSUE #1500: Adventure is still broken for 5.0.1
● **Describe the bug** 
  PacketEvents 2.12.1 throws ``NoClassDefFoundError``: ``net/kyori/adventure/util/Buildable$Builder`` when its update checker runs (and in any code path that loads GsonComponentSerializer$Builder). The bundled
  io.github.retrooper.packetevents.adventure.serializer.gson.GsonComponentSerializer$Builder interface still declares net.kyori.adventure.util.Buildable$Builder<...> as a superinterface, but that inner type was removed in Adventure 5.x (replaced by            
  net.kyori.adventure.builder.AbstractBuilder). The outer net.kyori.adventure.util.Buildable class still exists in Adventure 5.0.1, but the inner Builder interface does not, so any classloader resolution of GsonComponentSerializer$Builder fails when the
  runtime Adventure is 5.x.                                                                                                                                                                                                                                         
                       
  Verified with javap -v on packetevents-api-2.12.1.jar:
  public interface io.github.retrooper.packetevents.adventure.serializer.gson.GsonComponentSerializer$Builder
      extends net.kyori.adventure.builder.AbstractBuilder<...>,
              net.kyori.adventure.util.Buildable$Builder<...>,   // <-- missing in Adventure 5.x
              io.github.retrooper.packetevents.adventure.serializer.json.JSONComponentSerializer$Builder

  This is a binary incompatibility, not a relocation/shading issue — the same error reproduces whether PacketEvents is shaded or not, as long as the runtime Adventure is 5.x.

  ● **Software brand**
  Paper 1.21.11

  ● **Plugins**
  Only the plugin that bundles PacketEvents 2.12.1 alongside Adventure 5.0.1 needs to be present. No other plugins required to trigger it.

  Dependency set used in the failing build:
  - net.kyori:adventure-api:5.0.1
  - net.kyori:adventure-text-minimessage:5.0.1
  - net.kyori:adventure-text-serializer-gson:5.0.1
  - net.kyori:adventure-platform-bukkit:4.4.1
  - com.github.retrooper:packetevents-spigot:2.12.1

 ● **How To Reproduce**
  1. Create a plugin that depends on (or shades) packetevents-spigot:2.12.1 together with adventure-api:5.0.1 (and the matching text-serializer-gson:5.0.1).
  2. Initialize PacketEvents normally on plugin load (PacketEvents.getAPI().load() / init()), which kicks off the async update check.
  3. Start the server.
  4. Within a few seconds the packetevents-update-check-thread fails with the stack trace below.

  Reproduction does not require any client action — UpdateChecker.checkLatestReleasedVersion calls AdventureSerializer.gson(), which triggers loading of GsonComponentSerializer$Builder and immediately throws.

  ● **Expected behavior**
  PacketEvents 2.12.1 should be compatible with Adventure 5.x — its bundled adventure serializers should not declare net.kyori.adventure.util.Buildable$Builder as a superinterface, since that type no longer exists in Adventure 5.x. The AbstractBuilder
  superinterface alone is sufficient.

  ● **Additional context**
  Stack trace:
```
[13:42:15 WARN]: Exception in thread "packetevents-update-check-thread" java.lang.NoClassDefFoundError: to/itsme/itsmyconfig/shade/kyori/adventure/util/Buildable$Builder
[13:42:15 WARN]:        at java.base/java.lang.ClassLoader.defineClass1(Native Method)
[13:42:15 WARN]:        at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1027)
[13:42:15 WARN]:        at java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:150)
[13:42:15 WARN]:        at org.bukkit.plugin.java.PluginClassLoader.findClass(PluginClassLoader.java:263)
[13:42:15 WARN]:        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:593)
[13:42:15 WARN]:        at org.bukkit.plugin.java.PluginClassLoader.loadClass0(PluginClassLoader.java:180)
[13:42:15 WARN]:        at org.bukkit.plugin.java.PluginClassLoader.loadClass(PluginClassLoader.java:175)
[13:42:15 WARN]:        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:526)
[13:42:15 WARN]:        at itsmyconfig-4.3.2 (3).jar//to.itsme.itsmyconfig.shade.packetevents.util.adventure.AdventureSerializer.gson(AdventureSerializer.java:227)
[13:42:15 WARN]:        at itsmyconfig-4.3.2 (3).jar//to.itsme.itsmyconfig.shade.packetevents.util.updatechecker.UpdateChecker.checkLatestReleasedVersion(UpdateChecker.java:55)
[13:42:15 WARN]:        at itsmyconfig-4.3.2 (3).jar//to.itsme.itsmyconfig.shade.packetevents.util.updatechecker.UpdateChecker.checkForUpdate(UpdateChecker.java:70)
[13:42:15 WARN]:        at itsmyconfig-4.3.2 (3).jar//to.itsme.itsmyconfig.shade.packetevents.util.updatechecker.UpdateChecker.lambda$handleUpdateCheck$1(UpdateChecker.java:126)
[13:42:15 WARN]:        at java.base/java.lang.Thread.run(Thread.java:1583)
[13:42:15 WARN]: Caused by: java.lang.ClassNotFoundException: to.itsme.itsmyconfig.shade.kyori.adventure.util.Buildable$Builder
[13:42:15 WARN]:        at org.bukkit.plugin.java.PluginClassLoader.loadClass0(PluginClassLoader.java:208)
[13:42:15 WARN]:        at org.bukkit.plugin.java.PluginClassLoader.loadClass(PluginClassLoader.java:175)
[13:42:15 WARN]:        at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:526)
[13:42:15 WARN]:        ... 13 more
```
  Adventure 5.0.1 jar contents (relevant subset):
  net/kyori/adventure/util/Buildable.class            ← present
  net/kyori/adventure/util/Buildable$Builder.class    ← REMOVED in 5.x
  net/kyori/adventure/builder/AbstractBuilder.class   ← new replacement

booky10:
Reworks how packetevents handles adventure, should fix https://github.com/retrooper/packetevents/issues/1500

With this PR, packetevents now determines the currently installed version of adventure (if any) and downloads required dependencies with correct versions at runtime
This is not an optimal solution, though its the best solution in my opinion
I'd love to hear some feedback/other opinions on this

The options we currently have:
1. just stick with the current approach (shading without relocation) and work around all the weird issues this causes (not a good option)
2. relocate adventure (breaking change and is not an option for a library; there would be no way to "convert" between a packetevents `Component` class and the default adventure `Component` class, other than expensive serialization/deserialization)
3. drop support for platforms which either don't include adventure or include an outdated version of adventure (would be nice, though packetevents targets all versions starting with 1.8)
4. this PR (downloading adventure versions at runtime and injecting them into the platform classloader, won't work if offline; though e.g. paperclip also doesn't work if offline)

(this PR is also not fully done yet, we currently download adventure artifacts from maven central (against their ToS) and don't even cache them locally)


TheFaser:
That's a good way to solve problem, but I think relocate is the better solution because it's easier to support. I don't think deserialize/serialize causes any critical latency, but I haven't tested it. 

For example, I give my users a choice for this, and when they enable deserialize/serialize mode, they don't feel any message delay, even though my plugin sends messages through PacketEvents and it happens almost every tick. I haven't tested both cases, so I can't say for sure that this is a great way to do it, but I don't think it will be a bottleneck in PacketEvents.

Anyway, thank you for your work. You're doing a lot

booky10:
> For example, I give my users a choice for this, and when they enable deserialize/serialize mode, they don't feel any message delay, even though my plugin sends messages through PacketEvents and it happens almost every tick. I haven't tested both cases, so I can't say for sure that this is a great way to do it, but I don't think it will be a bottleneck in PacketEvents.

I get where you are coming from, but this still has the issue of breaking changes
And you would need to explicitely convert between net.kyori Component and a packetevents kyori Component on every method call which seems pretty excessive to me
I dont think this would be a good solution, especially for developer experience/ease of use

TheFaser:
> I get where you are coming from, but this still has the issue of breaking changes And you would need to explicitely convert between net.kyori Component and a packetevents kyori Component on every method call which seems pretty excessive to me I dont think this would be a good solution, especially for developer experience/ease of use

Developers will have to choose between using Kyori from PacketEvents or from Paper. In my personal experience with Paper, I very rarely need to use Kyori because most messages are sent through PacketEvents, which is more convenient.

Developers also have the option to relocate Kyori and PacketEvents libraries together to use their own version of Kyori.

I understand this is a breaking change, so it's probably best to discuss this for version 3.0.0. For now, your solution is good

rafi67000:
Im vouching for either 1 or 4

First option sucks for packevents but works.
Second option would be very annoying for developers.
Third option is a HUGE NO because PE's greatness comes from multiplatform and multiversion support.
Dropping versions with incompatible/no adventure would be a disaster for plugin maintainers that depend on packetevents.  
4th option is the best option in my opinion.

iiAhmedYT:
im sorry to question this but how would this fix #1500 ? it's a compile-time induced error not really dependent on the adventure version in the server

basically if packetevents is compiled with 4.26.1 you can't call any Serializer#builder on 5.0.1

if its compiled with 5.0.1 you can't call Serializer#builder in older versions

SamB440:
Would this approach fix the sponge problem? We shade adventure-nbt since sponge doesn't include it, but if two plugins shade it then everything breaks; so would this be checking if adventure-nbt exists and then download and install it if not?

booky10:
> im sorry to question this but how would this fix #1500 ?

@iiAhmedYT 
without this PR, we currently shade a "patched" version of adventure's gson serializer which targets 4.26.1 and thus causes #1500
https://github.com/retrooper/packetevents/blob/f7479fdf6dc12fa622f82b4d9805e18eafa906b9/patch/adventure-text-serializer-gson/src/main/java/net/kyori/adventure/text/serializer/gson/GsonComponentSerializer.java#L126

with this PR, we would no longer shade the adventure gson serializer ourselves but download it at runtime matching the adventure version currently installed on the platform (or defaulting to 4.26.1 if there is no adventure present)

feel free to test if this PR actually does fix the issue in your case, though it should work according to my testing

> Would this approach fix the sponge problem? We shade adventure-nbt since sponge doesn't include it, but if two plugins shade it then everything breaks; so would this be checking if adventure-nbt exists and then download and install it if not?

@SamB440 yes, although this PR currently only does this for spigot
integrating this into the sponge module should be pretty simple though

ytnoos:
Is removing Adventure from packetevents a valid alternative? Like completely stopping using it at all so packetevents doesn't need it anymore?

Or maybe distributing two different packetevents:

One with adventure bundled for legacy servers
The other without adventure, which relies on the one already present

booky10:
> Is removing Adventure from packetevents a valid alternative? Like completely stopping using it at all so packetevents doesn't need it anymore?

Would simplify things a lot, although you will still get breaking changes and less interopability with other plugins which may already use adventure

> One with adventure bundled for legacy servers
> The other without adventure, which relies on the one already present

This is basically what this PR does, but as one jar (and without hard failures)
Note that some parts of adventure aren't present on modern platforms - e.g. paper/sponge don't have adventure-nbt installed at all

ytnoos:
> > Is removing Adventure from packetevents a valid alternative? Like completely stopping using it at all so packetevents doesn't need it anymore?
> 
> Would simplify things a lot, although you will still get breaking changes and less interopability with other plugins which may already use adventure
> 
> > One with adventure bundled for legacy servers
> > The other without adventure, which relies on the one already present
> 
> This is basically what this PR does, but as one jar (and without hard failures)
> Note that some parts of adventure aren't present on modern platforms - e.g. paper/sponge don't have adventure-nbt installed at all

Then I think that implementing the local cache so the download is made just at first boot is 100% necessary. Also packetevents should check on boot if the hash of the downloaded libraries is the same as the one needed, in case you update packetevents which requires a different adventure version compared to the cached one. (maybe you have already done it, I haven't looked at diffs)

Regarding dropping it, we are skipping a few things because "it would break everything", but one day I hope packetevents will be going to switch major version without caring of breaking things

booky10:
> Then I think that implementing the local cache so the download is made just at first boot is 100% necessary. Also packetevents should check on boot if the hash of the downloaded libraries is the same as the one needed, in case you update packetevents which requires a different adventure version compared to the cached one. (maybe you have already done it, I haven't looked at diffs)

yes, this should be done 100% - this PR is not finished yet, just wanted to gather feedback on this approach and what others think about this

> Regarding dropping it, we are skipping a few things because "it would break everything", but one day I hope packetevents will be going to switch major version without caring of breaking things

v3 will contain breaking changes, though we aren't working on anything related to it yet

Rothes:
Do you think it's possible for a combination of 1 and 2?

Basically, add an middleware API, store raw component data, and lazy initialize adventure component; Or deserialize component back to NBT or json.
This way, the current API can be retained for Paper plugins (1), and make a copy of shaded adventure API too (2).
This can be achieved by moving the adventure related classes to a Gradle module and then creating another module to relocate tha module. Of course, making this module public and allowing developers to shade it on their plugin is also an option?
Developers can choose the adventure they want to use as needed and environment...

The reason I brought this up is because I have to use both server privided adventure and shaded adventure on my plugin as needed, on different packet types.

iiAhmedYT:
https://github.com/PaperMC/adventure/releases/tag/v5.2.0

@booky10 this might change things?

booky10:
> https://github.com/PaperMC/adventure/releases/tag/v5.2.0
> 
> @booky10 this might change things?

No, the release just fixes an ABI break which we already worked around in april

iiAhmedYT:
thanks for letting me know

booky10:
I'm not really sure if I'm still a fan of this PR though
We could never actually update to adventure api v5 for packetevents because of compatibility with legacy versions (like paper 1.16.5)

> > Is removing Adventure from packetevents a valid alternative? Like completely stopping using it at all so packetevents doesn't need it anymore?
> 
> Would simplify things a lot, although you will still get breaking changes and less interopability with other plugins which may already use adventure

I like this more and more, but I'm not sure whether this would be a feasible alternative

PR #1519, opened 2026-05-17
Fix team colors being erased if re-encoding WrapperPlayServerTeams on 1.8-1.12

### LINKED ISSUE #1517: Team colors get removed if re-encoding a WrapperPlayServerTeams
**Describe the bug**
When using WrapperPlayServerTeams, when it reads the team prefix and suffix, if it only contains color codes, it just writes an empty string back with no color codes, essentially making every team packet modification unable to do team colors

**Software brand**
Paper 1.8.8

**Plugins**
PacketEvents, WorldEdit, ViaVerison, My own custom one

**How To Reproduce**
1. Create a normal scoreboard team with the prefix being "§c§l" to make them have red bold name
2. Create a packet listener that uses WrapperPlayServerTeams, and modify anything, team color or just team name or something else
3. Every player on the team has a white name instead

**Expected behavior**
Modifying the WrapperPlayServerTeams should not be erasing colors

**Screenshots**
I ran a test using the same logic as the WrapperPlayServerTeams read/write methods, here are the results:

Image

Image

As you can see, even with this test, the color codes are gone when serialized back to a string, which is why team colors get erased

If you have actual characters in the prefix/suffix, then this does NOT happen, Adventure only does this if theres only color codes

LeonTG:
Closes #1517

### DIFF
--- a/patch/adventure-text-serializer-legacy/src/main/java/net/kyori/adventure/text/serializer/legacy/LegacyComponentSerializerImpl.java
+++ b/patch/adventure-text-serializer-legacy/src/main/java/net/kyori/adventure/text/serializer/legacy/LegacyComponentSerializerImpl.java
@@ -325,12 +325,12 @@ final class LegacyComponentSerializerImpl implements LegacyComponentSerializer {
 
     @Override
     public void component(final @NotNull String text) {
-      if (!text.isEmpty()) {
+//      if (!text.isEmpty()) { // packetevents comment out - scoreboard team prefix/suffix may have empty text but have color and formatting, which should be retained
         if (this.head < 0) throw new IllegalStateException("No style has been pushed!");
 
         this.styles[this.head].applyFormat();
         this.sb.append(text);
-      }
+//      } // packetevents comment out - scoreboard team prefix/suffix may have empty text but have color and formatting, which should be retained
     }
 
     @Override


booky10:
The adventure-patch modules will be removed eventually because we will replace our adventure handling with a more robust solution - though I don't know if there is a another good way to fix this, other than writing our own legacy component serializer

LeonTG:
I did try this too when testing it:

image

adding text there when serializing it and then removing the text after its been turned into text

but I found it kinda hacky solution

booky10:
We could also just passthrough plain text instead of using legacy component serialization at all

LeonTG:
You mean store it as string instead of component? wont that mess with 1.13+?

booky10:
I meant using a String for versions the protocol uses legacy strings and using a Component for versions the protocol uses components

TheFaser:
> adding text there when serializing it and then removing the text after its been turned into text

I have this problem too, and it's exactly MiniMessage issue (all serializers). And btw, I also solved this problem, just like you, by [adding a space and deleting it at the end](https://github.com/Flectone/FlectonePulse/blob/master/core/src/main/java/net/flectone/pulse/execution/pipeline/MessagePipeline.java#L92)

TheFaser:
MiniMessage adds a closing tag if there is no text after it, I think this is their usual behavior and they don't want to fix it

LeonTG:
> I meant using a String for versions the protocol uses legacy strings and using a Component for versions the protocol uses components

Maybe you guys should do that then cuz ngl I'm not entirely sure how you'd want me to structure that change with setting up 2 variables and how the setters/getters would then work

Depending on how long your solution will take tho, maybe this should be the solution in the meantime so this doesn't go unresolved for too long?

RaynLegends:
Adding my two cents since I've had this problem too.

This issue is caused by a bug in the legacy serializer (note that the other serializers do not have this behaviour, just the legacy one...) that could be fixed in Adventure upstream (https://github.com/PaperMC/adventure/pull/1312).

I think the best way other than getting that bug fixed (which is likely not happening) is a passthrough as suggested by booky10.

TheFaser:
> This issue is caused by a bug in the legacy serializer (note that the other serializers do not have this behaviour, just the legacy one...) that could be fixed in Adventure upstream

no, I have this issue in default MiniMessage serializer, and it's not solved by updating

RaynLegends:
> no, I have this issue in default MiniMessage serializer, and it's not solved by updating

Very weird, since the MiniMessage serializer does not strip trailing colors...

You can easily test this by trying:
```java
MiniMessage miniMessage = MiniMessage.miniMessage();

Component exampleComponent = text("").color(NamedTextColor.GOLD);
IO.println(miniMessage.serialize(exampleComponent));

String exampleString = "<yellow>";
IO.println(miniMessage.deserialize(exampleString));
```

Which produces the following:

image

Maybe this is a bit off topic though, happy to discuss on discord if you want (`@raynlegends`)

PR #1521, opened 2026-05-17
Consistency in WrapperPlayServerSoundEffect position

LeonTG:
**Justification:**

When working with positions in other wrappers, the position offset/multiplier is factored in to the constructor/setter/getter position variable

However the sound event was not doing this, which is a bit inconsistent both with the rest of this project and how the vanilla packet constructor/getter also functions

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerSoundEffect.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerSoundEffect.java
@@ -27,6 +27,7 @@ import com.github.retrooper.packetevents.protocol.sound.SoundCategory;
 import com.github.retrooper.packetevents.protocol.sound.Sounds;
 import com.github.retrooper.packetevents.protocol.sound.StaticSound;
 import com.github.retrooper.packetevents.resources.ResourceLocation;
+import com.github.retrooper.packetevents.util.Vector3d;
 import com.github.retrooper.packetevents.util.Vector3i;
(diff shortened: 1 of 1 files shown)


api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerSoundEffect.java
```diff
@@ -27,17 +27,19 @@
...
 import com.github.retrooper.packetevents.resources.ResourceLocation;
+import com.github.retrooper.packetevents.util.Vector3d;
 import com.github.retrooper.packetevents.util.Vector3i;
 import com.github.retrooper.packetevents.wrapper.PacketWrapper;
 import org.jetbrains.annotations.Nullable;
 
 import java.util.concurrent.ThreadLocalRandom;
 
 public class WrapperPlayServerSoundEffect extends PacketWrapper<WrapperPlayServerSoundEffect> {
+    private static final float POSITION_MULTIPLIER = 8.0F;
```
booky10, line 39: Any reason for not using double precision here? Vanilla uses doubles for writing and then casts to ints, which may make create a small difference in certain cases

LeonTG: Image

Vanilla packet has it as a float too, hence me copying that

However I just noticed yeah it casts it to double when multiplying by 8, so I pushed a commit now that fixes that

Image

it should now match vanilla exactly

booky10 (approved):

PR #1523, opened 2026-05-22
Add PluginMessage Common Wrapper

Rothes:
I'm thinking add another generic wrapper for both client and server common wrapper, since they have exactly same content.

What do you think? If so, we need to determine a package name for that, additionally, please share your thoughts on the parent class name. Personally, I believe the developer might be importing the wrong class if we add that. It might be confusing.


rafi67000:
everyone just used the existing plugin message because it's the same in every configuration phase

booky10 (approved):
I think its fine to leave clientbound/serverbound wrappers separated

PR #1526, opened 2026-05-26
Optimize 1.18+ chunk palette reading

linsaftw:
Improves chunk decode performance by replacing the slow NetStreamInput.readLong() palette path with direct PacketWrapper primitive reads. This avoids per-long byte-array allocation and removes the readBytes / copyMemory hotspot shown in Spark during chunk packet processing.

Testing phase in progress. Will report back when done.

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/protocol/stream/NetStreamInputWrapper.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/protocol/stream/NetStreamInputWrapper.java
@@ -33,6 +33,59 @@ public class NetStreamInputWrapper extends NetStreamInput {
     }
 
     @Override
+    public byte readByte() {
+        return this.wrapper.readByte();
+    }
+
+    @Override
+    public int readUnsignedByte() {
+        return this.wrapper.readUnsignedByte();
+    }
+
+    @Override
+    public short readShort() {
+        return this.wrapper.readShort();
+    }
+
+    @Override
+    public int readUnsignedShort() {
+        return this.wrapper.readUnsignedShort();
+    }
+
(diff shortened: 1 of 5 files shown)


purpurcof:
LGTM!!!

linsaftw:
Tested fully working and chunk writes CPU usage completely removed from spark.

Cyramek:
btw I've made experimental branch with chunk reader optimisations https://github.com/GrimAnticheat/packetevents/pull/12, if you want you can test that as well, I will deploy this to production later this week and see how it goes.

linsaftw:
> btw I've made experimental branch with chunk reader optimisations [GrimAnticheat#12](https://github.com/GrimAnticheat/packetevents/pull/12), if you want you can test that as well, I will deploy this to production later this week and see how it goes.

I took a very quick look at it, maybe then can be combined together?

Cyramek:
> I took a very quick look at it, maybe then can be combined together?

I did some more detailed benchmarks, and my solution for reading longs is actually usually slower than what you did. 
It’s faster in specific cases, but overall I wouldn’t merge that part.

There are a few things that could still be ported over, like the 1.7–1.8/1.9 optimisations and the ZeroBitStorage idea for empty chunks, but those would require a more thoughtful implementation. I’ll work on that in the coming days

booky10 (approved):
Refactored some methods into `PacketWrapper`, but looks fine to me otherwise

linsaftw:
Just noticed this! Thank you so much man!

booky10:
> Just noticed this! Thank you so much man!

Thank you for your PR! :heart:

PR #1446, opened 2026-01-25
Implement pre via version injection support for spigot module

ytnoos:
Since Retrooper gave us the opportunity to possibly accept the pre-via patch, here we are. I merged my patch with the fixes [[1]](https://github.com/GrimAnticheat/packetevents/commit/3684f01b032c522cab685718a6eb137845626f54) [[2]](https://github.com/GrimAnticheat/packetevents/commit/675bc5c8b321d043e863e41b55ca4f74d1c54a07) made by @Axionize, whom I thank very much for contributing and making this possible. Currently, an old version of this patch is used by both [Grim](https://github.com/GrimAnticheat/packetevents/) and [LPX](https://bstats.org/plugin/bukkit/LPX/9156).

By default, PacketEvents keeps the standard post-Via behavior. 
Pre-Via support is enabled dynamically when a Pre-Via listener is registered: in order to do that just override `isPreVia()` in your listener:
```
@Override
public boolean isPreVia() {
    return true;
}
```
I apologize in advance for any errors and look forward to hearing from you if any changes or improvements to the code are needed.

PR #1528, opened 2026-06-02
Instantly run task after write operation

Beaness:
Once again #1460 
> Right now it's possible other packets get written before the promise is completed, If you want to write a packet right after another packet using getTasksAfterSend it's possible another packet gets written inbetween. This solves it for spigot (sponge needs a different type of implementation, probably overwriting the whole write function to execute the tasks).

However this seems to have caused issues

> @Axionize (packetevents discord): 
https://github.com/GrimAnticheat/Grim/issues/2589
https://github.com/GrimAnticheat/Grim/issues/2588
https://github.com/GrimAnticheat/Grim/issues/2587
https://github.com/GrimAnticheat/Grim/issues/2586
https://github.com/GrimAnticheat/Grim/issues/2586
BadpacketsN also starts flagging
amongst many otherthings

Marked as draft until the issues have been fixed

### DIFF
--- a/fabric-common/src/main/java/io/github/retrooper/packetevents/handler/PacketDecoder.java
+++ b/fabric-common/src/main/java/io/github/retrooper/packetevents/handler/PacketDecoder.java
@@ -18,6 +18,9 @@
 
 package io.github.retrooper.packetevents.handler;
 
+import com.github.retrooper.packetevents.event.PacketSendEvent;
+import com.github.retrooper.packetevents.event.ProtocolPacketEvent;
+import com.github.retrooper.packetevents.exception.PacketProcessException;
 import com.github.retrooper.packetevents.protocol.PacketSide;
 import com.github.retrooper.packetevents.protocol.player.User;
 import com.github.retrooper.packetevents.util.PacketEventsImplHelper;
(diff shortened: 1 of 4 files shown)


Beaness:
No longer marked as draft, after further analysis it seems to be that the merge commit with Grim PacketEvents had inverted preVIa (https://github.com/GrimAnticheat/packetevents/commit/2f4163f#diff-3167a052813eba4c51f921f2e6d380977bd1de679ffb6e0bb62a1897d8c2c0bf)
Also the implementation of this w/ preVia requires extra logic to persist task order

```
tail
    -> normal PacketEvents   // sees packet first, registers/runs normal task
    -> ViaVersion
    -> preVia PacketEvents   // sees packet second, registers/runs preVia task
    -> head/socket

  Before PR:

  normal PE sees A       // adds normal task to promise first
  preVia PE sees A       // adds preVia task to same promise second
  promise completes
  normal task runs       // first registered
  preVia task runs       // second registered

  After PR:

  normal PE sees A
  normal PE calls ctx.write(A)
    -> ViaVersion
    -> preVia PE sees A
    -> preVia PE calls ctx.write(A)
    -> preVia task runs     // runs while nested inside normal ctx.write(A)
  normal task runs          // runs after ctx.write(A) returns to normal PE
 ```
 
A solution for this would be by instead of running the post tasks in preVia it adds them to a collection in normal pe handler so that order is persisted and the normal pe handler can run them after the write. But current PacketEvents does not implement preVia injection until https://github.com/retrooper/packetevents/pull/1446 is merged.

PR #1529, opened 2026-06-03
26.2 support

booky10:
Draft because 26.2 isn't released yet
PR is updated to 26.2-pre-3 (and should work with 26.2-pre-6 too), though I only did a short test on fabric


booky10:
Tested to work on both paper 26.2-rc-2 and fabric 26.2-rc-2, PR should be ready

booky10:
I will just merge this so dev builds include 26.2 support, if you find any issues with 26.2 please open an issue

PR #1530, opened 2026-06-05
Add nettystom (Minestom) platform module

TinyBrickBoy:
Integrate the nettystom platform from the packetstom fork into packetevents as an additional platform module alongside the existing ones (spigot, bungeecord, velocity, sponge, fabric).

nettystom targets a Netty-based Minestom fork and depends only on the shared :api and :netty-common modules. Registered in settings.gradle.kts without removing or altering any existing platform.


Tofaa2:
LFG AI SLOP

PR #1533, opened 2026-06-11
Gradle update

rafi67000:
Updated gradle, plugins and libraries for buildscript


booky10 (approved):
Seems fine, gson 2.14.0 aligns with vanilla 26.2 anyway

rafi67000:
it's only buildscript though you can parse the output in json format between different versions of gson and it will work perfectly

PR #1535, opened 2026-06-14
Annotate nullable manager returns

### LINKED ISSUE #1514: PlayerManager and ProtocolManager return nullable values but are not annotated with @Nullable
**Describe the bug**
PlayerManager and ProtocolManager return nullable values but are not annotated with `@Nullable`. This is really annoying and I've had a lot of bugs because of this plus my IDE keeps warning me about it when it is nullable.

**Software brand**
\

**Plugins**
\

**How To Reproduce**
\

**Expected behavior**
The [getChannel](https://github.com/retrooper/packetevents/blob/2fccee72cfceca10801046deb54d1a49112045fd/api/src/main/java/com/github/retrooper/packetevents/manager/player/PlayerManager.java#L35) and [getUser](https://github.com/retrooper/packetevents/blob/2fccee72cfceca10801046deb54d1a49112045fd/api/src/main/java/com/github/retrooper/packetevents/manager/player/PlayerManager.java#L37) method to be annotated as `@Nullable`

**Screenshots**
\

**Additional context**
\

Jan1k1:
Fixes #1514.

Adds nullable return annotations for manager lookups that can miss and return null.

Tested:
- .\gradlew.bat :api:test --console=plain

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/manager/player/PlayerManager.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/manager/player/PlayerManager.java
@@ -26,15 +26,16 @@ import com.github.retrooper.packetevents.protocol.player.User;
 import com.github.retrooper.packetevents.wrapper.PacketWrapper;
 import org.jetbrains.annotations.ApiStatus;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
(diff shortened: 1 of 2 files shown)


booky10 (approved):
Seems fine for now, I started to work a checkstyle implementation a few weeks ago which should catch + fix some more of these issues - but this will still take a while

PR #1537, opened 2026-06-20
fix: add type check for channel read handler

TheFaser:
Fixed `ClassCastException: class io.netty.channel.socket.DatagramPacket cannot be cast to class io.netty.channel.Channel`


booky10 (approved):
That was fast lol

booky10:
Thanks for the quick PR

Issue #1320, opened 2025-08-05
Add NeoForge Support


ir1skaa:
It's simple, add neoforge support for development on this mod loader as it is becoming more and more popular for modding

kenshineto:
I am the author of the following plugin: https://github.com/kenshineto/kenshinshideandseek2, which uses packet events. I would like to support neoforge in the future, so it would be great if this was implemented :)

Most of the net code that packet events uses is just `net.minecraft` so it should be (in theory ™️) pretty simple to implement

PR #1540, opened 2026-06-22
feat: add neoforge platform support

TheFaser:
#1320

This isn’t finished yet, because I have library issue on NeoForge. If you try to run packetevents-neoforge and adventure-platform-neoforge right now, you’ll get an error message. 

I also haven’t tested functionality and I’m not sure everything works at the moment, because (as I wrote above) I haven’t been able to resolve library conflict yet and because of that I can’t test PacketEvents together with my plugins. If you make the listeners directly in PacketEvents, that already works.

I currently have no ideas on how to solve this problem and I'm not insisting that NeoForge must be supported by PacketEvents but this platform is very similar to Fabric so supporting it isn't difficult.

UPD: I think the Adventure dependency issue also occurs on Fabric, so to solve it, we need to change loading this library, or update version to 5.x for NeoForge and Fabric


TheFaser:
I think this PR is ready. It might be worth testing it with other plugins that don't relate to PacketEvents, but I'm already using this PacketEvents on NeoForge for my project and there are no errors. Except for PacketEncoder, which I had to fix

PR #1543, opened 2026-06-28
fix: use correct NBT number types instead of always casting to int

TheFaser:


booky10 (approved):
Thanks for the quick fix ❤️

PR #1549, opened 2026-07-14
perf: drop per-call allocations in AdventureNBTSerializer style (de)serialization

R00tB33rMan:
Process + allocation spark (all corresponding "negligible" performance issues have been sufficiently resolved since this PR):
- Allocation: https://spark.lucko.me/vFIsYkEE5S
- Process: https://spark.lucko.me/LZlH1hqafl

`deserializeStyle` runs for every component and recurses through children, so it's on the hot path whenever text-heavy packets go through. A couple of things made it allocate more than it needs to:
- It iterated `TextDecoration.NAMES.keys()`, and Adventure's `Index#keys()` hands back a fresh `Set.copyOf(...)` on every call (same with `NAMES.values()` on the serialize side).
- Every style field was read via a `Consumer`/`Function` (`reader.useUTF(...)`, `useBoolean(...)`, ...), so each call spun up a pile of short-lived capturing lambdas; around 20 per component once you count the internal `useX -> useNumber -> useTag` delegates (x5 for the decorations); plus the invokedynamic linkage that comes with them.

This caches the decoration names/values in static arrays and adds small, lambda-free `getUTF`/`getNumber` readers, allowing the style attributes to be read directly instead of through a consumer. `serializeStyle` receives the same cached-decoration treatment. No behavior change; same NBT in/out.

The frames only showed up as ~0.02–0.05% in a profile, but they're hit recursively per nested component, so it compounds under component-heavy traffic (e.g., tab + scoreboards).

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/util/adventure/AdventureNBTSerializer.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/util/adventure/AdventureNBTSerializer.java
@@ -90,14 +90,7 @@ import java.util.function.Function;
 
 public class AdventureNBTSerializer implements ComponentSerializer<Component, Component, NBT> {
 
-    private static final TextDecoration[] DECORATION_VALUES = TextDecoration.values();
-    private static final String[] DECORATION_KEYS = new String[DECORATION_VALUES.length];
-
-    static {
-        for (int i = 0; i < DECORATION_VALUES.length; i++) {
-            DECORATION_KEYS[i] = DECORATION_VALUES[i].toString();
-        }
-    }
+    private static final Map<String, TextDecoration> DECORATION_MAP = new HashMap<>(TextDecoration.NAMES.keyToValue());
 
     private final ClientVersion version;
     private final boolean downsampleColor;
@@ -491,9 +484,9 @@ public class AdventureNBTSerializer implements ComponentSerializer<Component, Co
             if (shadowColor != null) style.shadowColor(ShadowColor.shadowColor(shadowColor.intValue()));
         }
 
-        for (int i = 0; i < DECORATION_KEYS.length; i++) {
-            Number value = reader.getNumber(DECORATION_KEYS[i]);
-            if (value != null) style.decoration(DECORATION_VALUES[i], TextDecoration.State.byBoolean(value.byteValue() != 0));
+        for (Map.Entry<String, TextDecoration> entry : DECORATION_MAP.entrySet()) {
+            Number value = reader.getNumber(entry.getKey());
+            if (value != null) style.decoration(entry.getValue(), TextDecoration.State.byBoolean(value.byteValue() != 0));
         }
 
         String insertion = reader.getUTF("insertion");
@@ -617,10 +610,10 @@ public class AdventureNBTSerializer implements ComponentSerializer<Component, Co
             if (shadowColor != null) writer.writeInt("shadow_color", shadowColor.value());
         }
 
-        for (TextDecoration decoration : DECORATION_VALUES) {
-            TextDecoration.State state = style.decoration(decoration);
+        for (Map.Entry<String, TextDecoration> entry : DECORATION_MAP.entrySet()) {
+            TextDecoration.State state = style.decoration(entry.getValue());
             if (state != TextDecoration.State.NOT_SET) {
-                writer.writeBoolean(decoration.toString(), state == TextDecoration.State.TRUE);
+                writer.writeBoolean(entry.getKey(), state == TextDecoration.State.TRUE);
             }
         }
 
--- a/api/src/main/java/com/github/retrooper/packetevents/util/adventure/AdventureNBTSerializer.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/util/adventure/AdventureNBTSerializer.java
@@ -88,9 +88,16 @@ import java.util.Optional;
 import java.util.function.Consumer;
 import java.util.function.Function;
 
-import static com.github.retrooper.packetevents.util.adventure.AdventureIndexUtil.indexValueOrThrow;
-
 public class AdventureNBTSerializer implements ComponentSerializer<Component, Component, NBT> {
+
+    private static final TextDecoration[] DECORATION_VALUES = TextDecoration.values();
+    private static final String[] DECORATION_KEYS = new String[DECORATION_VALUES.length];
+
+    static {
+        for (int i = 0; i < DECORATION_VALUES.length; i++) {
+            DECORATION_KEYS[i] = DECORATION_VALUES[i].toString();
+        }
+    }
 
     private final ClientVersion version;
     private final boolean downsampleColor;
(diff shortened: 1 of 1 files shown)


R00tB33rMan (commented):
Since I cannot scope change requests, my only existing concern is converting this from a `decoration.toString()`. I'm not entirely sure if this is guaranteed to be the same.

I considered looping over `DECORATION_MAP.values()` (which doesn't create a copy, e.g. no performance impact and use the old `decoration.toString()`); however, I'd like to leave this at the discretion of the maintainer.

`writer.writeBoolean(entry.getKey(), state == TextDecoration.State.TRUE);` (Modified code section)

Tofaa2:
How are you making commits while not knowing java??

R00tB33rMan:
> How are you making commits while not knowing java??

I've been learning! I feel like this is a somewhat misplaced assumption. I made sure to test these changes as well. Let's keep this PR on-topic. I apologize if this comes off as rude or negatively construed.

Tofaa2:
> > How are you making commits while not knowing java??
> 
> I've been learning! I feel like this is a somewhat misplaced assumption. I made sure to test these changes as well. Let's keep this PR on-topic. I apologize if this comes off as rude or negatively construed.

Consumer callbacks on the jvm with JIT are essentially a free operation. I dont see why make the code less readable for absolutely no benefit. Lambdas get JIT optimized so i have a feeling you ai'ed this.

Also the refactoring of your new methods into this absolutely destroys any chance of branch prediction happening. The only thing I can get behind is Adventure returning a copy for their API instead of ah immutable view but that to me seems like an adventure PR

WouterGritter:
> > > How are you making commits while not knowing java??
> > 
> > 
> > I've been learning! I feel like this is a somewhat misplaced assumption. I made sure to test these changes as well. Let's keep this PR on-topic. I apologize if this comes off as rude or negatively construed.
> 
> Consumer callbacks on the jvm with JIT are essentially a free operation. I dont see why make the code less readable for absolutely no benefit. Lambdas get JIT optimized so i have a feeling you ai'ed this.
> 
> Also the refactoring of your new methods into this absolutely destroys any chance of branch prediction happening. The only thing I can get behind is Adventure returning a copy for their API instead of ah immutable view but that to me seems like an adventure PR

See the referenced memory allocation spark! Magnolia > Netty threads. This method is being hammered recursively. Object allocations for each lambda _do_ add up, and more than half of the optimization here is avoiding calling `TextDecoration.NAMES.keys()` on the hot path, which was creating a mutable copy of the keyset on every call, something that's not needed here.

XIII-MC:
buzzword buzzword buzzword

Bram1903:
> buzzword buzzword buzzword

What's the point of saying that under a PR?

api/src/main/java/com/github/retrooper/packetevents/util/adventure/AdventureNBTSerializer.java
```diff
@@ -470,22 +470,27 @@ public AdventureNBTSerializer(boolean downsampleColor) {
...
+            if (shadowColor != null) style.shadowColor(ShadowColor.shadowColor(shadowColor.intValue()));
         }
 
-        for (String decorationKey : TextDecoration.NAMES.keys()) {
-            reader.useBoolean(decorationKey, value -> style.decoration(
-                    indexValueOrThrow(TextDecoration.NAMES, decorationKey),
-                    TextDecoration.State.byBoolean(value)));
+        for (Map.Entry<String, TextDecoration> entry : DECORATION_MAP.entrySet()) {
+            Number value = reader.getNumber(entry.getKey());
+            if (value != null) style.decoration(entry.getValue(), TextDecoration.State.byBoolean(value.byteValue() != 0));
```
booky10: A new method reader.getBoolean would make this more descriptive I think

api/src/main/java/com/github/retrooper/packetevents/util/adventure/AdventureNBTSerializer.java
```diff
@@ -88,10 +88,10 @@
 import java.util.function.Consumer;
 import java.util.function.Function;
 
-import static com.github.retrooper.packetevents.util.adventure.AdventureIndexUtil.indexValueOrThrow;
-
 public class AdventureNBTSerializer implements ComponentSerializer<Component, Component, NBT> {
 
+    private static final Map<String, TextDecoration> DECORATION_MAP = new HashMap<>(TextDecoration.NAMES.keyToValue());
```
booky10: Index.keyToValue only exists since adventure v4.10.0, you will probably need to manually build this map (or add a util method to AdventureIndexUtil)

api/src/main/java/com/github/retrooper/packetevents/util/adventure/AdventureNBTSerializer.java
```diff
@@ -868,6 +873,18 @@ public <R> R readUTF(String key, Function<String, R> function) {
             return withTag(key, tag -> function.apply(requireType(tag, NBTType.STRING).getValue()));
         }
 
+        public String getUTF(String key) {
+            NBT tag = compound.getTagOrNull(key);
+            return tag == null ? null : requireType(tag, NBTType.STRING).getValue();
+        }
+
+        public Number getNumber(String key) {
```
booky10:
```suggestion
        public @Nullable Number getNumber(String key) {
```

api/src/main/java/com/github/retrooper/packetevents/util/adventure/AdventureNBTSerializer.java
```diff
@@ -868,6 +873,18 @@ public <R> R readUTF(String key, Function<String, R> function) {
             return withTag(key, tag -> function.apply(requireType(tag, NBTType.STRING).getValue()));
         }
 
+        public String getUTF(String key) {
```
booky10:
```suggestion
        public @Nullable String getUTF(String key) {
```

api/src/main/java/com/github/retrooper/packetevents/util/adventure/AdventureNBTSerializer.java
booky10: You could also just store TextDecoration.NAMES.values() as a field instead, as you can easily get the decoration key from the value using TextDecoration.toString

booky10 (changes requested):
Minor changes, though I can't edit your PR for some reason

booky10:
> > > How are you making commits while not knowing java??
> > 
> > 
> > I've been learning! I feel like this is a somewhat misplaced assumption. I made sure to test these changes as well. Let's keep this PR on-topic. I apologize if this comes off as rude or negatively construed.
> 
> Consumer callbacks on the jvm with JIT are essentially a free operation. I dont see why make the code less readable for absolutely no benefit. Lambdas get JIT optimized so i have a feeling you ai'ed this.
> 
> Also the refactoring of your new methods into this absolutely destroys any chance of branch prediction happening. The only thing I can get behind is Adventure returning a copy for their API instead of ah immutable view but that to me seems like an adventure PR

I don't think its that bad to get rid of nested lambda calls

R00tB33rMan:
> Minor changes, though I can't edit your PR for some reason

Ah, yeah, likely since this is an organization. I'll make these adjustments for you!

api/src/main/java/com/github/retrooper/packetevents/util/adventure/AdventureNBTSerializer.java
R00tB33rMan: Yeah, this looks better as well!

R00tB33rMan:
Hey @booky10, anything else you need me to do with this PR, or just lettin' it sit for a while until you have time/motivation?

booky10 (approved):

booky10:
Thanks for the optimization and the PR!

PR #1557, opened 2026-07-28
fix: replace obsolete PaintingType with PaintingVariant in WrapperPlayServerSpawnPainting

### LINKED ISSUE #1556: WrapperPlayServerSpawnPainting takes PaintingType (obsolete) without alternatives
## Describe the bug
[`WrapperPlayServerSpawnPainting`](https://github.com/retrooper/packetevents/blob/3e0e4a2573fbdec79863580d1c09937363c13bae/api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerSpawnPainting.java#L34) takes [`PaintingType`](https://github.com/retrooper/packetevents/blob/2.0/api/src/main/java/com/github/retrooper/packetevents/protocol/world/PaintingType.java#L28) which is marked as obsolete, without any alternatives.

## Software brand
/

## Plugins
Please give us a list of your plugins. Sometimes we run into incompatibilities. Having an overview of the plugins on your server can really help identify the problem quicker.
Eg. ViaVersion, ProtocolLib, HolographicDisplays
You may also provide a screenshot.
/

## How To Reproduce
*Example:*
1. Join with the 1.18.2 vanilla minecraft client.
2. Spam
3. ....
/

## Expected behavior
/

bejiihiu:
Fixes #1556.

## Problem
`WrapperPlayServerSpawnPainting` used `@ApiStatus.Obsolete` `PaintingType` enum for **all** server versions without supporting the modern `PaintingVariant` registry-based system.

Since 1.19+, Minecraft sends the painting type as a `painting_variant` registry ID in the Spawn Painting packet, which does **not** match the old enum ordinals. For example, `donkey_kong` moved from ID 25 (1.18.2) to ID 29 (1.19+), and `earth`/`wind`/`water`/`fire` shifted by one position. The old code would read/write wrong variants on 1.19+.

Additionally, the 1.21 painting update added 20+ new variants (sorted alphabetically, completely different ID order) that the old enum cannot represent.

## Fix
- Replaced the internal field from `@Nullable PaintingType type` to `@Nullable PaintingVariant variant`
- **1.13+**: Delegates to `readMappedEntity`/`writeMappedEntity` using the `PaintingVariants` registry, which has per-version ID mappings from `painting_variant.json` (covers 1.7.2, 1.19, 1.21, 1.21.7)
- **Pre-1.13**: Converts the PascalCase string title (e.g. `"SkullAndRoses"`) to the snake_case registry key (`"skull_and_roses"`) via a regex split on camelCase boundaries

## Backward Compatibility
All deprecated `PaintingType` API preserved:
- `@Deprecated` constructors accepting `PaintingType` convert to `PaintingVariant` internally
- `@Deprecated` `getType()` returns `Optional<PaintingType>` by matching the variant's registry key back to the enum
- `@Deprecated` `setType(PaintingType)` converts via enum-name lookup
- Helpers handle the `PIG_SCENE` → `"pigscene"` special case (enum name has underscore but registry key does not)

## New API
```java
WrapperPlayServerSpawnPainting(int entityId, UUID uuid, @Nullable PaintingVariant variant, Vector3i position, Direction direction)
PaintingVariant getVariant()
void setVariant(PaintingVariant)
```

## Verification
- `:api:compileJava` — **BUILD SUCCESSFUL** (0 new warnings)
- `:api:test` — **all tests pass**

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerSpawnPainting.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerSpawnPainting.java
@@ -88,9 +88,8 @@ public class WrapperPlayServerSpawnPainting extends PacketWrapper<WrapperPlaySer
         if (serverVersion.isNewerThanOrEquals(ServerVersion.V_1_13)) {
             this.variant = readMappedEntity(PaintingVariants.getRegistry());
         } else {
-            String title = readString(13);
-            this.variant = PaintingVariants.getByName(
-                    title.replaceAll("([a-z])([A-Z])", "$1_$2").toLowerCase());
+            PaintingType oldType = PaintingType.getByTitle(readString(13));
+            this.variant = oldType != null ? paintingTypeToVariant(oldType) : null;
         }
         if (serverVersion.isNewerThanOrEquals(ServerVersion.V_1_8)) {
             this.position = readBlockPosition();
@@ -116,7 +115,8 @@ public class WrapperPlayServerSpawnPainting extends PacketWrapper<WrapperPlaySer
         if (serverVersion.isNewerThanOrEquals(ServerVersion.V_1_13)) {
             writeMappedEntity(Objects.requireNonNull(this.variant, "variant must be set"));
         } else {
-            writeString(pascalCaseTitle(Objects.requireNonNull(this.variant, "variant must be set").getName().getKey()), 13);
+            PaintingVariant v = Objects.requireNonNull(this.variant, "variant must be set");
+            writeString(variantToPaintingType(v).getTitle(), 13);
         }
         if (serverVersion.isNewerThanOrEquals(ServerVersion.V_1_8)) {
             long positionVector = this.position.getSerializedPosition(this.serverVersion);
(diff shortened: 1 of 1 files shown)


bejiihiu:
Updated:

- `write()` now uses `Objects.requireNonNull(variant, "variant must be set")` instead of silently NPE-ing — gives a clear message if someone forgets to set the variant before writing
- `variantToPaintingType()` replaced O(n) linear search with a `HashMap` built once in `static {}` — constant-time lookup

api/src/main/java/com/github/retrooper/packetevents/wrapper/play/server/WrapperPlayServerSpawnPainting.java
```diff
@@ -68,9 +86,11 @@ public void read() {
             this.uuid = new UUID(0L, 0L);
         }
         if (serverVersion.isNewerThanOrEquals(ServerVersion.V_1_13)) {
-            this.type = PaintingType.getById(readVarInt());
+            this.variant = readMappedEntity(PaintingVariants.getRegistry());
         } else {
-            this.type = PaintingType.getByTitle(readString(13));
+            String title = readString(13);
+            this.variant = PaintingVariants.getByName(
+                    title.replaceAll("([a-z])([A-Z])", "$1_$2").toLowerCase());
```
NoJokeFNA: Is this regular expression really needed?

bejiihiu: Good point, the regex was overkill. Replaced it — pre-1.13 now uses `PaintingType.getByTitle()` (which already maps PascalCase titles correctly) then converts through `paintingTypeToVariant()`. Same for writing: `variantToPaintingType(v).getTitle()`. Much simpler.

(sorry for English, I use a translator)

Jan1k1:
Superseded by https://github.com/retrooper/packetevents/pull/1584 — same idea, with a safe pre-1.13 write for modern-only variants and nullable getVariant().

PR #1558, opened 2026-07-28
refactor(api): rename EntityTypes fields for naming consistency

bejiihiu:
## Problem

The `EntityTypes` class had inconsistent naming that made version-agnostic entity mapping unnecessarily hard (#1552):

- `LIVINGENTITY` — missing underscore separator (should be `LIVING_ENTITY`)
- `PROJECTILE_ABSTRACT` — `ABSTRACT_` prefix convention violated (should be `ABSTRACT_PROJECTILE`)
- `MINECART_ABSTRACT` — same convention issue (should be `ABSTRACT_MINECART`)
- `CHESTED_MINECART_ABSTRACT` — same convention issue (should be `ABSTRACT_CHESTED_MINECART`)
- `BOAT` extended `ENTITY` directly — no abstract parent type for boat hierarchy

This forced downstream mappers (e.g. [PacketEntities' EntityTypeMapper](https://github.com/3add/PacketEntities/blob/master/code-generators/src/main/java/dev/threeadd/packetentities/codegen/data/mapping/EntityTypeMapper.java)) to maintain manual override tables because a simple `CamelCase → UPPER_SNAKE_CASE` conversion produced names that didn't exist.

## Changes

| Old (inconsistent) | New (canonical) |
|---|---|
| `LIVINGENTITY` | `LIVING_ENTITY` |
| `PROJECTILE_ABSTRACT` | `ABSTRACT_PROJECTILE` |
| `MINECART_ABSTRACT` | `ABSTRACT_MINECART` |
| `CHESTED_MINECART_ABSTRACT` | `ABSTRACT_CHESTED_MINECART` |
| *(missing)* | `ABSTRACT_BOAT` |
| *(missing)* | `ABSTRACT_CHEST_BOAT` |

- Old constants are preserved as `@Deprecated` aliases pointing to the new canonical names — zero breakage for existing API consumers
- `BOAT` now extends `ABSTRACT_BOAT` and `CHEST_BOAT` now extends `ABSTRACT_CHEST_BOAT` instead of `ENTITY`/`BOAT`
- All internal hierarchy references updated to use canonical names

## Backward Compatibility

All four old constant names (`LIVINGENTITY`, `PROJECTILE_ABSTRACT`, `MINECART_ABSTRACT`, `CHESTED_MINECART_ABSTRACT`) remain accessible with `@Deprecated` javadoc pointing to their replacements. No existing code using these constants needs to change — the deprecated fields compile and resolve to the same registry entries.

## Verification

[...]

PR #1559, opened 2026-07-28
feat(api): split EntityTypes hierarchy into EntityClass + EntityHierarchy

### LINKED ISSUE #1552: EntityTypes class is not well designed
## Describe the bug
The current state of the EntityTypes class is not very great for version agnostic purposes, it misses classes and its naming is inconsistent. If you'd like to see more issues, here's a remapper I had to implement [PacketEntities' EntityTypeMapper.java](https://github.com/3add/PacketEntities/blob/master/code-generators/src/main/java/dev/threeadd/packetentities/codegen/data/mapping/EntityTypeMapper.java).

## Software brand
``` 
Eg. Spigot 1.18.2, Paper 1.18.2, Tuinity 1.18.2, Pufferfish 1.18.2, etc...
Make sure to mention the version of the software.
```
/, issue exists in the main `api` module.

## Plugins
```
Please give us a list of your plugins. Sometimes we run into incompatibilities. Having an overview of the plugins on your server can really help identify the problem quicker.
Eg. ViaVersion, ProtocolLib, HolographicDisplays
You may also provide a screenshot.
```
/, issue exists in the main `api` module.

## How To Reproduce
```
Steps to reproduce the behavior:
*Example:*
1. Join with the 1.18.2 vanilla minecraft client.
2. Spam
3. ....
4. See error
```
/, issue exists in the main `api` module.

## Expected behavior
```
A clear and concise description of what you expected to happen.
```
A better implementation

## Screenshots
```
If applicable, add screenshots to help explain your problem.
```

## Additional context
```
Add any other context about the problem here.
```

bejiihiu:
## Problem

`EntityTypes` mixed two completely separate concepts in one class:

1. **The `minecraft:entity_type` registry** — ~158 concrete, spawnable types (zombie, creeper, pig...)
2. **The entity class hierarchy** — abstract parent types (LIVING_ENTITY, ABSTRACT_MONSTER...) that exist only for `isInstanceOf` checks

Abstract types were stored in the same `VersionedRegistry` with `id=-1` for every version. Passing one to a spawn packet would silently write `-1` into the protocol. `EntityTypes.values()` returned ~200 entries when only ~158 are real registry types.

## Changes

| Concept | Before | After |
|---|---|---|
| Abstract types location | In `EntityTypes` + `VersionedRegistry` | **New `EntityHierarchy` class** — no registry involved |
| Hierarchy node type | `EntityType` (`id=-1`, `isRegistered()=true`) | **New `EntityClass`** — pure tree node, no protocol IDs |
| Concrete type parent type | `@Nullable EntityType` | **`EntityClass`** — structurally impossible to pass an abstract type |
| `values()` | ~200 (abstract + concrete) | **~158 (concrete only)** |
| `isInstanceOf` API | `(EntityType)` only | **`(EntityClass)`** primary; `(EntityType)` deprecated |

## Design

Inspired by ViaVersion's entity type system (`EntityType` interface with `isAbstractType()`, `identifier()`, `isOrHasParent()`). Adapted to packetevents' `VersionedRegistry` architecture with the addition of `EntityClass` — a standalone immutable tree node that mirrors Minecraft's actual class hierarchy.

```
EntityHierarchy (EntityClass tree)         EntityTypes (VersionedRegistry)
  ENTITY                                     ZOMBIE = define("zombie", ABSTRACT_MONSTER)
  LIVING_ENTITY                              CREEPER = define("creeper", ABSTRACT_MONSTER)
  ABSTRACT_MONSTER                           DROWNED = define("drowned", ABSTRACT_MONSTER, ZOMBIE)
  ...                                         ...
        ↓                                           |
        └─── EntityType.isInstanceOf(EntityClass) ←─┘
```

## Backward Compatibility

All old `EntityTypes.*` hierarchy constants (`LIVING_ENTITY`, `ABSTRACT_MONSTER`, etc.) are preserved as `@Deprecated` bridge types. Old code compiling against `EntityTypes.LIVING_ENTITY` still compiles unchanged. `EntityTypes.isTypeInstanceOf()` still works via the deprecated `isInstanceOf(EntityType)` method.

## Bug Fixes

- **Fixed** `INTERACTION` parent from `DISPLAY` → `ENTITY` (Interaction extends Entity directly in Minecraft, not Display)
- `EntityClass.getName()` / `getParent()` follow JavaBean convention matching the rest of the codebase
- `isAssignableFrom(null)` throws `NullPointerException` as `java.lang.Class` does

## Verification

- `:api:compileJava` — BUILD SUCCESSFUL (0 new warnings)
- `:api:test` — all tests pass
- Full project `compileJava` — BUILD SUCCESSFUL

Closes #1552

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/protocol/entity/type/EntityClass.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/protocol/entity/type/EntityClass.java
@@ -1,0 +1,79 @@
+/*
+ * This file is part of packetevents - https://github.com/retrooper/packetevents
+ * Copyright (C) 2026 retrooper and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.github.retrooper.packetevents.protocol.entity.type;
+
+import java.util.Objects;
+
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * A node in the Minecraft entity class hierarchy. Mirrors the Java inheritance
+ * structure of Minecraft's entity classes (Entity → LivingEntity → Mob → ...).
+ * <p>
+ * Unlike {@link EntityType}, this is NOT a registry entry. It carries no
+ * protocol
+ * IDs, no versioned mappings, and is not stored in any registry. It exists
+ * purely
+ * for type-checking via {@link EntityType#isInstanceOf(EntityClass)}.
+ * <p>
+ * Design inspired by ViaVersion's entity type system.
+ *
+ * @see EntityHierarchy All predefined hierarchy nodes
+ */
+public final class EntityClass {
+
+	private final String name;
+	private final @Nullable EntityClass parent;
+
+	EntityClass(String name, @Nullable EntityClass parent) {
+		this.name = name;
+		this.parent = parent;
+	}
+
+	public String getName() {
+		return this.name;
+	}
+
+	public @Nullable EntityClass getParent() {
+		return this.parent;
+	}
+
+	/**
+	 * Returns true if {@code possibleChild} is this class or a descendant.
+	 * Walks the parent chain and compares by reference equality.
+	 *
+	 * @throws NullPointerException if possibleChild is null
+	 * @see Class#isAssignableFrom(Class)
+	 */
+	public boolean isAssignableFrom(EntityClass possibleChild) {
+		Objects.requireNonNull(possibleChild, "possibleChild");
+		EntityClass current = possibleChild;
+		while (current != null) {
+			if (current == this)
+				return true;
+			current = current.parent;
+		}
+		return false;
+	}
+
+	@Override
+	public String toString() {
+		return "EntityClass{" + name + "}";
+	}
+}
--- a/api/src/main/java/com/github/retrooper/packetevents/protocol/entity/type/EntityHierarchy.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/protocol/entity/type/EntityHierarchy.java
@@ -1,0 +1,111 @@
+/*
+ * This file is part of packetevents - https://github.com/retrooper/packetevents
+ * Copyright (C) 2026 retrooper and contributors
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+package com.github.retrooper.packetevents.protocol.entity.type;
+
+/**
+ * Predefined {@link EntityClass} hierarchy nodes mirroring Minecraft's
+ * entity class inheritance tree. These are NOT registry entries — they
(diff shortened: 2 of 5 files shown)


bejiihiu:
Thanks to @3add for the detailed feedback on #1558 — the suggestions about separating the registry from the hierarchy directly shaped this design. I was initially worried this would take a very long time, but after looking at how ViaVersion and similar projects already implement entity type hierarchies, it turned out to be surprisingly clean and fast to adapt. Hope this meets expectations!

booky10:
Could you please disclose which part of this was human written and which part was written by an LLM?

3add:
The hard part about representing this hierarchy is that it should be version agnostic. I assume the tree in this PR is based on 26.2? Booky and I discussed generating these hierarchies based on data generators that fetched their data from decompiled mc. That way it would be less tedious to create. But I still don't exactly know what the best way would be to implement it (version specific classes?).

bejiihiu:
> Не могли бы вы раскрыть, какая часть написана человеком, а какую — LLM?

I mostly use AI to help write my PR descriptions since my English isn't the best. I also sometimes use it while coding to sanity-check things, find hacky parts of my code, or point out questionable solutions. It's mainly there to make my PRs easier to read and review, and I don't think it takes away from the actual problem-solving.

bejiihiu:
Actually, the tree isn't specifically based on 26.2 — it's the same hierarchy that already existed in the old `EntityTypes` class (originally credited to ViaVersion [here](https://github.com/retrooper/packetevents/blob/3e0e4a2573fbdec79863580d1c09937363c13bae/api/src/main/java/com/github/retrooper/packetevents/protocol/entity/type/EntityTypes.java#L72)), just extracted into its own file. The hierarchy is designed to be version-agnostic: Minecraft's Java class inheritance (Entity → LivingEntity → Mob → Monster → Zombie, etc.) is stable across versions. What changes between versions is which concrete types exist and what protocol IDs they have — that's handled by `entity_type.json`, not by the hierarchy.

Version-specific classes (like ViaVersion's per-version enums) are one approach, but I think they'd be overkill here. PacketEvents' strength is the JSON-based mapping system — one hierarchy + per-version ID mappings means you don't need to ship a new enum every release. New mobs just get assigned to an existing EntityClass node, and the JSON file handles the rest.

Auto-generating the hierarchy from decompiled MC sounds like an interesting long-term idea, especially for catching new abstract types Mojang adds. But the current tree is only ~33 nodes and changes maybe once a year, so manual maintenance has been working well so far.

Could we chat on Telegram or Discord to discuss a slightly different approach to this PR?\

3add:
> Minecraft's Java class inheritance (Entity → LivingEntity → Mob → Monster → Zombie, etc.) is stable across versions. 

This is simple untrue, there are many abstract classes that get added/removed throughout the versions, just as an example you have `Boat extends Entity` in 1.14.4 but in 26.2 it's `Boat extends AbstractBoat extends Entity`. See https://kennytv.eu/entity-data/?diff=&from=1.14.4

bejiihiu:
You're right about the versioning, but the static hierarchy isn't something this PR introduced. Here's the original line from `origin/2.0`:

```java
public static final EntityType BOAT = define("boat", ENTITY); // no ABSTRACT_BOAT at all
```

There was never any ABSTRACT_BOAT — BOAT was a direct child of ENTITY for all versions, even though in 26.2 it should be Boat → AbstractBoat → Entity. This PR at least adds ABSTRACT_BOAT and improves the situation. The hierarchy was also originally credited to ViaVersion and was always static.

This PR just moves the existing hierarchy into its own class — it doesn't make the versioning problem worse, it makes it easier to fix later.

3add:
> This PR just moves the existing hierarchy into its own class — it doesn't make the versioning problem worse, it makes it easier to fix later.

Why PR something that has to be fixed later when you can do it right from the first time, this PR doesn't fully fix the issue, it moves it to another class which makes it so we have to deprecate more later when we do fix it. It should be done right from the start.

3add:
I suggest one of two things for the hierarchy: either have version specific classes with static typed values or have version specific jsons with the full hierarchy representation, each node gets a static counterpart (for all vers) and then the `instanceOf` check just looks through the version specific json.

bejiihiu:
Versioned JSON hierarchy for 33 nodes across 21 versions? That's ~700 lines of `"monster": "pathfinder_mob"` boilerplate per version, a whole new JSON mapping system to maintain, and a registry loader for something that currently works with zero reported bugs in the 5+ years the old EntityTypes has been in production.

Sure, I can do it. It's just code. But I think there are 100 more impactful things packetevents could use that development time on.

bejiihiu:
Sorry, I'm not skilled enough to implement versioned hierarchy properly. I'll close this PR and delete my IDE. Thanks for your time.

3add:
💀, like I said the json's would be auto generated. Packetevents also has a similar mapping system already implemented for their mapped type stuff so it's not a crazy suggestion to do the same for entity types.

bejiihiu:
Hope someone picks this up. Feel free to use my code as a base or inspiration if you do.

thx

PR #1562, opened 2026-08-02
fix 1.7 interact action

ManInMyVan:


rafi67000:
oh so that's why interactions were cooked on 1.7 LMAO

booky10 (approved):
Thanks for the fix

PR #1564, opened 2026-08-05
Use unpooled buffers for item stack conversion

RareMan666:
Temporary ItemStack conversions used pooled direct buffers, causing Netty arenas to retain memory after release.
Using unpooled buffers avoids this unnecessary native memory growth.

### DIFF
--- a/spigot/src/main/java/io/github/retrooper/packetevents/util/SpigotReflectionUtil.java
+++ b/spigot/src/main/java/io/github/retrooper/packetevents/util/SpigotReflectionUtil.java
@@ -34,7 +34,7 @@ import com.github.retrooper.packetevents.util.reflection.ReflectionObject;
 import com.github.retrooper.packetevents.wrapper.PacketWrapper;
 import com.google.common.collect.BiMap;
 import com.google.common.collect.MapMaker;
-import io.netty.buffer.PooledByteBufAllocator;
+import io.netty.buffer.Unpooled;
 import org.bukkit.Bukkit;
 import org.bukkit.Material;
 import org.bukkit.NamespacedKey;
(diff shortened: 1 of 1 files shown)


booky10:
We explicitly release the buffer after usage and return its memory back to the pool

How did you come to the conclusion that we leak memory here? Did you measure any leaks caused by this?

RareMan666:
> We explicitly release the buffer after usage and return its memory back to the pool
> 
> How did you come to the conclusion that we leak memory here? Did you measure any leaks caused by this?

I found this through native-memory profiling on a Minecraft server running GrimAC

  I first collected long-running jemalloc heap profiles then used async-profiler nativemem together with JVM NMT. The dominant retained
  allocation stack was:

  Grim TickInventory
  → CorrectingPlayerInventoryStorage
  → BukkitPlatformInventory.getStack
  → PacketEvents.fromBukkitItemStack
  → SpigotReflectionUtil.decodeBukkitItemStack
  → PooledByteBufAllocator.DEFAULT.buffer
  → Unsafe_AllocateMemory0

The stack grew from approximately 24 MiB to 76 MiB, while NMT Other increased from 64 MiB to 153 MiB. I am not claiming that the buffer is missing a release(). The observation is that this temporary conversion path causes pooled direct-memory chunks to remain retained by Netty arenas

booky10:
Sounds expected, pooled memory is kept active in the pool so it doesn't need to be allocated again if we create a new buffer
Have you done a native memory profile with your unpooled changes? If possible, try to write a small program which just converts item stacks a few times, so you can directly compare the allocations

RareMan666:
> Sounds expected, pooled memory is kept active in the pool so it doesn't need to be allocated again if we create a new buffer Have you done a native memory profile with your unpooled changes? If possible, try to write a small program which just converts item stacks a few times, so you can directly compare the allocations

after thinking for a while, I think your explanation is correct, the retained memory appears to be netty pooled allocator keeping its chunks active for reuse, rather than an unreleased bytebuf

PR #1566, opened 2026-08-05
fix: PluginMessage behaves differently from Vanilla

ChengZhiMeow:


booky10 (approved):
Only affects clientbound packets, so I don't see any issues with increasing this cap
Thanks for your PR!

PR #1570, opened 2026-08-24
docs: added documentation to the bundle packet

Bram1903:
Recently I have seen an increase in the usage of the bundle packet, as well as people struggling to use it correctly. This adds a documentation block on top of the bundle wrapper explaining what the delimiter does, and how to send one without breaking a bundle somebody else already opened.


retrooper:
Thanks for the PR.

PR #1571, opened 2026-08-25
fix block action type id

### LINKED ISSUE #1392: WrapperPlayServerBlockAction returns wrong block when used on chests
**Describe the bug**
The WrapperPlayServerBlockAction#getBlockType() method returns stripped dark oak log instead of chest

**Software brand**
Paper 1.21.8, packetevents 2.10.0

**Plugins**
None except for my plugin

**How To Reproduce**
Listen to WrapperPlayServerBlockAction
Open a chest
Log the result of wrapper.blockType

**Expected behavior**
It should return a chest state, but it returns stripped dark oak log

**Screenshots**

```kotlin
override fun onPacketSend(event: PacketSendEvent) {
        val type = event.packetType

        if (type != PacketType.Play.Server.BLOCK_ACTION) return
        val wrapper = WrapperPlayServerBlockAction(event)

        val block = SpigotConversionUtil.toBukkitBlockData(wrapper.blockType)

        Bukkit.broadcast("block $block")
        if (block !is Chest) return
        Bukkit.broadcast("chest")
}
```

Jan1k1:
Fixes #1392

BLOCK_ACTION sends a block type id, not a state id.
getBlockType() was looking it up in the state table so chests came out as random blocks.


booky10 (approved):
Thanks for the fix!

PR #1572, opened 2026-08-25
dont inline simple items in ItemStack.encode before 26.1

### LINKED ISSUE #1569: Invalid ItemDialogBody encoding in some minecraft versions / cases
**Describe the bug**
In minecraft versions from **1.20.5** to **26.1-snapshot-3**, the `item` tag of an [ItemDialogBody](https://github.com/retrooper/packetevents/blob/2.0/api/src/main/java/com/github/retrooper/packetevents/protocol/dialog/body/ItemDialogBody.java) is encoded incorrectly when the item stack is considered to be "simple" (does not contain any component patches). This causes the client to disconnect with a Network Protocol Error once the body is used in a dialog and the client receives a `show_dialog` packet.

Here is a part of the stacktrace the vanilla client will provide along with the protocol error (using a normal Written Book item in this case):

```java
Caused by: io.netty.handler.codec.DecoderException: Failed to decode: Not a map: "minecraft:writable_book"
```

This implies that the client expects the item data as a "map" (nbt compound), but only the item ID is being provided as a regular string. Taking a look at the ItemDialogBody and ItemStack classes, the stack is encoded using a deprecated method called `ItemStack#encodeForParticle`. That method checks if the stack has any patched components, and if not, returns a `NBTString` instead of a compound:

https://github.com/retrooper/packetevents/blob/f52b3b92323b380f2ffb061ebd8e22ef1836d75c/api/src/main/java/com/github/retrooper/packetevents/protocol/item/ItemStack.java#L144-L152

The interesting part is that this behavior seems to work fine in newer versions of the game, like 26.1.2. I checked the vanilla sources for the decoding of the item dialog body, and found that in **26.1-snapshot-3**, this line in `net.minecraft.server.dialog.body.ItemBody` changes from:

```java
ItemStack.STRICT_CODEC.fieldOf("item").forGetter(ItemBody::item)
```

to instead:

```java
ItemStackTemplate.CODEC.fieldOf("item").forGetter(ItemBody::item)
```

The only difference here is that the codec changes, from `ItemStack.STRICT_CODEC` (which is mostly a reference to `ItemStack.MAP_CODEC` and therefore requires the data to be formatted as a map/compound) to a new codec specified in `ItemStackTemplate` (which appears to be new in that snapshot). That codec explicitly supports "inline" item references like the one PacketEvents is trying to insert, meaning the encoding is valid again there.

As the name says, `#encodeForParticle` is only used in one other place to encode additional particle data. In versions prior to that snapshot, an almost identical version of the new codec from above was defined and used *exclusively* in `net.minecraft.core.particles.ItemParticleOption`, and I assume the version logic used by PE was originally designed to work with that part only.

This leaves a window in between `1.20.5` (from the version check) and `26.1-snapshot-3`, in which the game refuses any "inline" item ID strings in the dialog body.

**Software brand**
Probably not relevant, the only important part should be the server version, as mentioned above.

**Plugins**
This shouldn't really be relevant, but this bug can be reproduced on the latest version of packetevents (2.13.0) on an otherwise empty paper server, as long as all of the criteria above is met.

**How To Reproduce**
1. Create a simple test dialog with an item body, without any modified components *(see example code below)*
2. Run the test on an affected version from **1.20.5** to **26.1-snapshot-3**
3. Show the dialog to the player (they will disconnect due to an exception)
4. Modify the item stack code part and add any component
5. The dialog will show fine now *(tho the component won't actually apply, since that method has a **TODO** for actually encoding the components)*

<details>
<summary>My testing code</summary>

```java
ItemStack stack = new ItemStack.Builder()
                .type(ItemTypes.DIAMOND_SWORD)
                .build();
DialogBody body = new ItemDialogBody(stack, null, true, true, 30, 30);
Dialog dialog = new NoticeDialog(new CommonDialogData(
        text("testing"), null, true, false,
        DialogAction.CLOSE, List.of(body), List.of()
), new ActionButton(new CommonButtonData(text("yes"), null, 100), null));
getCommand("test").setExecutor((sender, command, label, args) -> {
    if (!(sender instanceof Player p)) return false;
    WrapperPlayServerShowDialog wrapper = new WrapperPlayServerShowDialog(dialog);
    PacketEvents.getAPI().getPlayerManager().sendPacket(p, wrapper);
    return true;
});

```
</details>

Jan1k1:
Fixes #1569

ItemStack.encode was using the particle path, which inlines simple items as a string from 1.20.5.
Dialogs still want a compound until 26.1, so the client kicked with Not a map.

### DIFF
--- a/api/src/main/java/com/github/retrooper/packetevents/protocol/item/ItemStack.java
+++ b/api/src/main/java/com/github/retrooper/packetevents/protocol/item/ItemStack.java
@@ -137,7 +137,16 @@ public class ItemStack {
     }
 
     public static NBT encode(PacketWrapper<?> wrapper, ItemStack itemStack) {
-        return encodeForParticle(itemStack, wrapper.getServerVersion().toClientVersion());
+        ClientVersion version = wrapper.getServerVersion().toClientVersion();
+        NBT nbt = encodeForParticle(itemStack, version);
+        if (version.isNewerThanOrEquals(ClientVersion.V_1_20_5)
+                && version.isOlderThan(ClientVersion.V_26_1)
+                && nbt instanceof NBTString) {
+            NBTCompound compound = new NBTCompound();
+            compound.setTag("id", nbt);
+            return compound;
+        }
+        return nbt;
     }
 
     @Deprecated


booky10:
Thanks for the PR and sorry for the wait
I decided to re-implement ItemStack NBT-based serialization entirely, and properly separate ItemStack and ItemStackTemplate serialization
See https://github.com/retrooper/packetevents/commit/2e49ec0965d949de69f54f6f2eea5be75ca25ac0 for more details

PR #1585, opened 2026-09-15
Itemstack encoding and decoding optimizations

AbhigyaKrishna:
Optimized some of the hot path of nms <---> pe itemstack. Also introduced caching.


spigot/src/main/java/io/github/retrooper/packetevents/util/SpigotReflectionUtil.java
```diff
@@ -886,33 +902,183 @@ public static Object createNMSItemStack(Object nmsItem, int count) {
...
+
+    public static Material getMaterialByItemType(ItemType itemType) {
+        Material material = ITEM_TYPE_TO_MATERIAL.get(itemType);
+        if (material == null) {
+            material = encodeBukkitItemStackSlow(com.github.retrooper.packetevents.protocol.item.ItemStack
+                    .builder().type(itemType).build()).getType();
+            ITEM_TYPE_TO_MATERIAL.put(itemType, material);
+        }
+        return material;
+    }
```
booky10: Wouldn't a registry lookup make much more sense on modern versions? See https://jd.papermc.io/paper/26.3/org/bukkit/Registry.html#ITEM and https://jd.papermc.io/paper/26.3/org/bukkit/Registry.html#MATERIAL

PR #1586, opened 2026-09-18
Fix legacy particle duplicate ID lookup (#1581)

### LINKED ISSUE #1581: Silently thrown exception in 1.8.8
**Describe the bug**
When placing a block on a plant block, PacketEvents throws a Exception silently.
The Problem is, that if you write an ChannelHandler which catches Exception through the exceptionCaught() method, the silent exception also get's caught.

**Software brand**
PandaSpigot 1.8.8

**Plugins**
ViaVersion, LuckPerms, labymod-server-api-bukkit

**How To Reproduce**
*Example:*
1. Join the server
2. Place grass on a grass block (the plant grass)
3. get kicked

**Expected behavior**
That the exception does not throw.

Jan1k1:
Pre-1.13 particle mappings reuse one modern name for several EnumParticle ids (e.g. block = BLOCK_CRACK + BLOCK_DUST). TypesBuilder only kept the last id, so getById(37) was null and re-encoding NPEd on 1.8.8.

Register every id that shares a name, and preserve the raw type id on the particle wrapper for honest round-trips.

Closes #1581


jstindev:
Thank you

PR #1587, opened 2026-09-20
Fix goat horn default instrument on 26.x

### LINKED ISSUE #1563: Instrument component isn't serialized when packetevents is loaded
**Describe the bug**
On 26.2, goat horns sent to the client are missing the instrument component for specific instruments such as admire, this causes the dreaded itemstack hash mismatch which causes the item to rubberband in the inventory.

**Software brand**
Folia 26.2

**Version**
packetevents-spigot-2.13.0 from modrinth

**Plugins**
Packetevents

**How To Reproduce**
/give minecraft:goat_horn[instrument="minecraft:admire_goat_horn"]
For some reason it only breaks with admire
Shows on client side as ponder

Packet events loaded:
Serverside nbt: minecraft:goat_horn[instrument="minecraft:admire_goat_horn"]
Clientside nbt: {count:1,id:"minecraft:goat_horn"} (missing instrument)

Packet events unloaded:
Serverside nbt: minecraft:goat_horn[instrument="minecraft:admire_goat_horn"]
Clientside nbt: {components:{"minecraft:instrument":"minecraft:admire_goat_horn"},count:1,id:"minecraft:goat_horn"}

**Expected behavior**
Horn nbt should not change when passing through packetevents
In fact, no nbt should change simply by being deserialized by packet events, there are probably other instances of this that I have not discovered, some way to validate packetevents isnt mutating nbt would be a good addition

Jan1k1:
Fixes wrong base-component default for goat horn instrument on 26.1–26.3 (was admire / holder id 5, should be ponder).

Closes #1563


booky10:
These are auto-generated from vanilla registries (see https://github.com/packetevents/data-generators)
This issue should be fixed there instead

Jan1k1:
Moved the fix to the generators: https://github.com/packetevents/data-generators/pull/1

Hand-editing instrument.json here would just get overwritten on the next generate.

PR #1588, opened 2026-09-22
Mark getPlayer as NullUnmarked for Kotlin nullable casts

### LINKED ISSUE #1516: Kotlin/JSpecify: PlayerEvent#getPlayer<Player?>() fails despite @UnknownNullability because T is @NullMarked non-null
**Describe the bug**

`PlayerEvent#getPlayer()` / `ProtocolPacketEvent#getPlayer()` is annotated as returning `@UnknownNullability T`, and the backing `player` field/constructor parameter are also `@UnknownNullability`:

- `PlayerEvent#getPlayer()`: https://github.com/retrooper/packetevents/blob/2.0/api/src/main/java/com/github/retrooper/packetevents/event/PlayerEvent.java#L30-L38
- `ProtocolPacketEvent#player`: https://github.com/retrooper/packetevents/blob/2.0/api/src/main/java/com/github/retrooper/packetevents/event/ProtocolPacketEvent.java#L45-L57
- `ProtocolPacketEvent#getPlayer()`: https://github.com/retrooper/packetevents/blob/2.0/api/src/main/java/com/github/retrooper/packetevents/event/ProtocolPacketEvent.java#L155-L157

Those annotations were introduced in commit `548312ea34fe5b823be42c69d0a892fd06242bf0` (`Fix nullability annotations for packet events`):

https://github.com/retrooper/packetevents/commit/548312ea34fe5b823be42c69d0a892fd06242bf0

Because `PlayerEvent` and `ProtocolPacketEvent` are annotated with `@NullMarked`, Kotlin treats the generic type parameter as having a non-null upper bound (`T : Any`). As a result, Kotlin consumers cannot call `getPlayer<Player?>()` and null-check the result, even though the player can be null.

Minimal Kotlin call site:

```kotlin
val player = event.getPlayer<Player?>() ?: return
```

Compiler error with Kotlin `2.3.20`:

```text
Type argument is not within its bounds: must be subtype of 'Any'.
```

The published bytecode for PacketEvents `2.12.1` is effectively:

```text
public <T extends java.lang.Object> T getPlayer();
Signature: <T:Ljava/lang/Object;>()TT;
```

with `@UnknownNullability` only on the method return, not on the generic type parameter bound. Under `@NullMarked`, Kotlin sees `T` as non-null-bounded, so `Player?` is not a valid type argument.

**Software brand**

This is a compile-time API interoperability issue, not a server runtime issue.

Observed in a Paper plugin project using:

- PacketEvents API/Spigot: `2.12.1`
- Kotlin: `2.3.20`
- JVM toolchain: `25`
- Paper API target in the project: `26.1.2.build.63-stable`

**Plugins**

Not runtime/plugin-list dependent. The issue reproduces at Kotlin compile time with `packetevents-api` and a Bukkit/Paper API on the compile classpath.

**How To Reproduce**

1. Create a Kotlin JVM project using PacketEvents `2.12.1`, Kotlin `2.3.20`, and any Bukkit/Paper API that provides `org.bukkit.entity.Player`.
2. Add this single Kotlin file:

```kotlin
import com.github.retrooper.packetevents.event.PacketReceiveEvent
import org.bukkit.entity.Player

object PacketEventsNullabilityRepro {
    fun handle(event: PacketReceiveEvent) {
        val player: Player = event.getPlayer<Player?>() ?: return
        println(player.uniqueId)
    }
}
```

3. Compile.
4. Kotlin reports:

```text
Type argument is not within its bounds: must be subtype of 'Any'.
```

**Expected behavior**

Kotlin consumers should be able to represent the actual nullable player case and safely null-check it, for example:

```kotlin
val player = event.getPlayer<Player?>() ?: return
```

Given that PacketEvents marks the field, constructor parameter, and return value with `@UnknownNullability`, the generic signature should not force a non-null type argument in Kotlin.

**Screenshots**

N/A. This is a compiler error.

**Additional context**

Relevant PacketEvents source snippets are linked here instead of copied inline:

- `PlayerEvent#getPlayer()`: https://github.com/retrooper/packetevents/blob/2.0/api/src/main/java/com/github/retrooper/packetevents/event/PlayerEvent.java#L30-L38
- `ProtocolPacketEvent#player`: https://github.com/retrooper/packetevents/blob/2.0/api/src/main/java/com/github/retrooper/packetevents/event/ProtocolPacketEvent.java#L45-L57
- `ProtocolPacketEvent#getPlayer()`: https://github.com/retrooper/packetevents/blob/2.0/api/src/main/java/com/github/retrooper/packetevents/event/ProtocolPacketEvent.java#L155-L157

The regression appears to have started in `2.11.2`. The relevant commit is included in `v2.11.2`, `v2.12.0`, and `v2.12.1`:

- Commit: https://github.com/retrooper/packetevents/commit/548312ea34fe5b823be42c69d0a892fd06242bf0
- Compare: https://github.com/retrooper/packetevents/compare/v2.11.1...v2.11.2.diff

Possible fixes to consider:

1. Make the type parameter bound explicitly nullable under JSpecify, e.g. conceptually:

```java
<T extends @Nullable Object> @UnknownNullability T getPlayer();
```

2. Mark this specific API/scope as `@NullUnmarked` if the generic method is intended to preserve platform/unknown nullability for Kotlin consumers.

The main issue is not whether every event always has a null or non-null player; it is that the current combination of `@NullMarked` + generic `<T>` + `@UnknownNullability` return makes Kotlin reject nullable type arguments while the API contract and runtime behavior allow the player to be null.

Jan1k1:
Fixes #1516.

`@NullMarked` on `PlayerEvent` made Kotlin reject `getPlayer<Player?>()` even though the player can be null. Marking `getPlayer` `@NullUnmarked` (as booky suggested) keeps the existing `@UnknownNullability` contract while allowing nullable type arguments.


AbhigyaKrishna:
Thank you for the pr!

